PR #8, opened 2015-01-03
Fixed deprecated syntax

Detegr:
[0, ..1024] syntax is now deprecated. Fixed using std::iter::repeat.

I tried out `Vec::with_capacity` instead and passing `buffer.capacity()` instead of `buffer.len()` to get rid of initializing the vec with zeros, but got a failed assertion from slice.rs, so I guess the initialization is necessary.

### DIFF
--- a/src/wrapper.rs
+++ b/src/wrapper.rs
@@ -104,7 +104,7 @@ impl INotify {
 	pub fn available_events(&mut self) -> IoResult<&[Event]> {
 		self.events.clear();
 
-		let mut buffer = [0u8, ..1024];
+		let mut buffer = [0u8; 1024];
 		let len = unsafe {
 			ffi::read(
 				self.fd,


passcod:
My build output mentions that `[0, ..1]` syntax is now experimental, not deprecated, and recommends:

```help: add #![feature(slicing_syntax)] to the crate attributes to enable```

…where did you see that it's deprecated?

**Comment by Detegr** [comment]
I might be wrong of course, but slicing syntax is a different thing if I understood correctly. It's for an immutable slice to a vec using [2..4] style syntax. [See here](http://doc.rust-lang.org/std/slice/#traits).

There's now [a test in Rust repository](https://github.com/rust-lang/rust/blob/13392d19cae2b03883d531a47d660b985b98833b/src/test/compile-fail/array-old-syntax-1.rs) to test that this syntax produces an error, so I figured that that syntax is going away.

**Comment by passcod** [comment]
Here's a more complete error message for the record:

```
/home/.cargo/registry/src/github.com-1ecc6299db9ec823/inotify-0.1.0/src/wrapper.rs:107:26: 107:32 error: range syntax is experimental
/home/.cargo/registry/src/github.com-1ecc6299db9ec823/inotify-0.1.0/src/wrapper.rs:107          let mut buffer = [0u8, ..1024];
                                                                                                                       ^~~~~~
/home/.cargo/registry/src/github.com-1ecc6299db9ec823/inotify-0.1.0/src/wrapper.rs:107:26: 107:32 help: add #![feature(slicing_syntax)] to the crate attributes to enable
/home/.cargo/registry/src/github.com-1ecc6299db9ec823/inotify-0.1.0/src/wrapper.rs:107          let mut buffer = [0u8, ..1024];
```

This mentions using `#![feature(slicing_syntax)]` for `error: range syntax is experimental`. So these are indeed the same!

As for the test, I'm not sure, but it might be that it's to test the behaviour we're seeing, without a connotation of deprecation. I'll ask in IRC and report back.

passcod:
Nevermind, it's right there in the IRC topic: `[T, ..n] is now [T; n] | ...` So there you have it :)

Detegr:
Great that you found that out. I updated the PR to change the syntax to [T; n] form. I think it's more clear than std::iter::repeat.

passcod:
While on the topic of deprecated syntax, we should probably change `#[deriving(` to `#[derive(`.

hannobraun:
There's a build failure on Travis, but that seems unrelated. I tested locally and confirmed that this works with my nightly (from 3 days ago).

Thanks @Detegr for the pull request, and thanks @passcod for the help in sorting this out!

hannobraun:
FYI, I fixed the Travis build failure and also the `deriving` deprecation warnings. Everything works on the latest nightly :)

PR #10, opened 2015-01-09
Fix c_str redesign errors

dead10ck:
* Fixes the compilation errors caused by [the recent `c_str` redesign](https://github.com/rust-lang/rust/pull/20507).
* [`int`/`uint` have been renamed to `isize`/`usize`](https://github.com/rust-lang/rfcs/pull/544#issuecomment-68760871)
* Changes tabs to be 4 spaces per the [guideline](http://aturon.github.io/style/whitespace.html).
* Fixes a few compiler warnings, including using the new slicing syntax.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /Cargo.lock
 /target
 *.sublime-workspace
+*.swp
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -31,9 +31,9 @@
 //! [inotify(7)]: http://man7.org/linux/man-pages/man7/inotify.7.html
 
 use libc::{
-	c_char,
-	c_int,
-	uint32_t };
+    c_char,
+    c_int,
+    uint32_t };
 
 
 pub use libc::close;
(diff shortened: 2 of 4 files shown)


hannobraun:
Aside from the `offset(1)` thing, which I'm unsure about, all looks well! Thank you, @dead10ck!

The offset thing seems to be used for the name of the file that's changed. None of the automated tests cover this functionality. I think the best solution would be to add a test for this, so we can be sure the code was right. As before, I will get to it on Monday, or you can take a look, if you want.

Thanks again for all your work!

hannobraun:
Regarding the unstable warnings, I've opened a separate issue: #11.

dead10ck:
@hannobraun yeah, the offset thing should be checked before this is merged, but I won't have time to do it this weekend :/ and no problem!

hannobraun:
@dead10ck Ok, I'll pick it up on Monday then :)

hannobraun:
Merged manually. The `offset(1)` was indeed needed. The reason is that the `inotify_event` struct originally contains an optional string as its last field. Since I don't know how to represent that in Rust (not sure it can be), this requires some trickery when reading the name. See the [man page](http://linux.die.net/man/7/inotify) for all the details.

Thanks again, @dead10ck, for the great work!

dead10ck:
Ahh, I see. Sorry, I'm still new to Rust, and I haven't worked with lower-level FFI stuff yet, so I wasn't familiar with how all of that worked. I'm glad I was able to help with the other stuff though :)

hannobraun:
No need to apologize, that part of the code is a bit icky. We have a test for it now, so this problem shouldn't come up any more.

PR #12, opened 2015-02-01
Fixed build error (std::io => std::old_io)

paholg:
std::io has moved to std::old_io, as outlined here:

http://internals.rust-lang.org/t/psa-io-old-io/1403


hannobraun:
Thank you!

PR #13, opened 2015-02-05
Update for std::path reform in rustc version ba2f13ef0

daboross:
This updates for the changes to `std::path` added in  https://github.com/rust-lang/rust/pull/21759. (rfc: https://github.com/rust-lang/rfcs/blob/master/text/0474-path-reform.md) 

- Switches to using `unix::OsStrExt` (`path.as_os_str().as_byte_slice()`) for conversion from path to CString
- Switches to using `String::from_utf8()` (`String::from_utf8(c_str.as_bytes().to_vec())`) for conversion from CString to String
- Removes path feature, as it is no longer required

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,7 +2,7 @@
 #![crate_type = "lib"]
 #![warn(missing_docs)]
 #![stable]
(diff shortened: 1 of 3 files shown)


hannobraun:
Excellent, thank you!

PR #17, opened 2015-02-20
to_cstring() returns a Result now

dead10ck:
* `to_cstring()` returns a `Result` now, so compilation was failing
* `CString::from_slice()` is being deprecated, so I updated this to `CString::new()`
* Update syntax `[]` to `[..]`
* In the tests, use the new `File` and `Path`s. AFAICT, `temp_dir()` still returns an `old_path::Path`, and there isn't a newer `temp_dir()` method anywhere that returns a new `Path` yet.


hannobraun:
Looks good, thank you!

PR #18, opened 2015-03-07
Add example to REAME

GuillaumeGomez:
I like to have a little example like that on the README. Always helpful !

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,49 @@ Include it in your Cargo.toml:
 inotify = "*"
 ```
 
+And here's a little example:
+```Rust
+#![feature(path)]
+
+extern crate inotify;
+
+use inotify::INotify;
+use inotify::ffi::*;
+use std::path::Path;
+
+fn main() {
+    let mut ino = INotify::init().unwrap();
+
+    ino.add_watch(Path::new("/home"), IN_MODIFY | IN_CREATE | IN_DELETE).unwrap();
+    loop {
+        let events = ino.wait_for_events().unwrap();
+
+        for event in events.iter() {
+            if event.mask & IN_CREATE > 0 {
+                if event.mask & IN_ISDIR > 0 {
+                    println!("The directory \"{}\" was created.", event.name);       
+                } else {
+                    println!("The file \"{}\" was created.", event.name);
+                }
+            } else if event.mask & IN_DELETE > 0 {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,6 +24,49 @@ Include it in your Cargo.toml:
...
+
+fn main() {
+    let mut ino = INotify::init().unwrap();
+
+    ino.add_watch(Path::new("/home"), IN_MODIFY | IN_CREATE | IN_DELETE).unwrap();
+    loop {
+        let events = ino.wait_for_events().unwrap();
+
+        for event in events.iter() {
+            if event.mask & IN_CREATE > 0 {
```
hannobraun: This can be replaced by `event.is_create()`. The same goes for a few similar lines below. See [the methods on `Event`](https://github.com/hannobraun/inotify-rs/blob/2e6ab84acd8df2a9480af00b0d61fc9d672988c1/src/wrapper.rs#L216) for what's available.

hannobraun:
Thank you, this is awesome!

I've add a comment about a few lines that could be improved, in my opinion. If you don't have time to correct those, just say so. I'm happy to merge as is and improve the example later.

README.md
GuillaumeGomez: I only converted C code to Rust but these functions are very nice ! I change that.

GuillaumeGomez:
Done !

README.md
```diff
@@ -24,6 +24,49 @@ Include it in your Cargo.toml:
...
+
+fn main() {
+    let mut ino = INotify::init().unwrap();
+
+    ino.add_watch(Path::new("/home"), IN_MODIFY | IN_CREATE | IN_DELETE).unwrap();
+    loop {
+        let events = ino.wait_for_events().unwrap();
+
+        for event in events.iter() {
+            if event.is_create() > 0 {
```
hannobraun: Now it's wrong :)
`is_create` (and the other `Event` functions) return bool, so the `> 0` won't compile.

GuillaumeGomez: Totally ! My bad, I misread...

GuillaumeGomez:
I fixed comparison. Sorry for my misreading. ^^'

hannobraun:
Don't feel sorry, it was an honest mistake :)

Thanks again, merging!

PR #19, opened 2015-03-19
OsStrExt has moved, update import path. Use cheaper travis servers.

neuschaefer:
… cheaper in the sense that the main build servers are often overbooked, while the container-based build infrastructure mostly isn't (http://www.traviscistatus.com/).


hannobraun:
Thank you!

PR #21, opened 2015-03-19
fix path to OsStrExt

blaenk:


hannobraun:
This change has already been made in #19. Still, thanks a lot!

PR #22, opened 2015-04-04
Update to remove all usages of unstable features

daboross:
In `src`:
- Replaces dependency on in-rustc libc with the libc crate on crates.io
- Removes unrequired imports of `std::ffi::AsOsStr`, as Path implements a separate `.as_os_str()` method without the trait
- Replaces usage of `OsStr.to_cstring()` with `CString::new()`.
  - Note: this replaces a match statement with `try!()`
    - `io::Error` implements `From<NulError>`, so we can just use `try!()` instead of manually crafting an `io::Error`.
- Replaces usage of `slice.position_elem()` with `.splitn()`
  - This usage of `.nsplit(|b| b == &0u8).next().unwrap()` splits the slice on the first `\0` byte, taking everything before it, or just leaves it as it is if the slice doesn't have any `\0` bytes
    - Note: the `2` passed to `splitn()` ensures that `splitn` won't continue checking the string for more `\0` bytes after it has found one `\0` byte.
  - This exactly mirrors the behavior of the code it replaced, it just uses .split() instead of manually checking for the first instance of `\0` and splitting the slice manually.

In `tests`:
- Replaces usage of `file.path()` with just using the `PathBuf` produced by `temp_file()`.
  - As far as I can tell, both file.path() and the PathBuf produced should be exactly the same, as the `PathBuf` is literally taken from the exact file which was opened.
- Switches usage of `PathBuf::from(path.as_os_str())` to `PathBuf::from(path)`

This also updates `.travis.yml` to build and test with both the 1.0.0 beta build, and the latest nightly (instead of just the latest nightly).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: rust
 
+# Test using both the rust 1.0.0-beta, and the latest nightly
+rust:
+  - 1.0.0-beta
+  - nightly
+
 # Run this build on the "container-based infrastructure"
 # See http://docs.travis-ci.com/user/workers/container-based-infrastructure/.
 sudo: false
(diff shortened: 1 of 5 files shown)


hannobraun:
Thanks a lot for the contribution, and sorry for the delay.

daboross:
No problem, thank you!

PR #23, opened 2015-04-25
removed stability attributes

tsurai:
A recent Rust nightly update removed the #[stable], #[unstable] and #[deprecated] attributes for non-#[staged_api] crates (rust-lang/rust#22830) in favor of a more complex release system ([RFC](https://github.com/rust-lang/rfcs/blob/master/text/0507-release-channels.md)).

Stability attributes are now bound to features, mainly intended for Rusts own release cycle. I suggest removing the previously used stability attributes since inotify is already stable enough and not frequently changing as if it would benefit much from it.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -1,5 +1,3 @@
-#![stable]
-
 //! Bindings for inotify
 //!
 //! There are four types of statics:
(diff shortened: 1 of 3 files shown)


hannobraun:
Merged. Thank you!

Issue #28, opened 2015-10-29
wrapper abstraction does not lend itself to multithreading


benschulz:
When one thread reads from the inotify fd and another closes it, bad things can happen.

 - The reading thread gets stuck in `wait_for_events` forever.
 - The inotify fd becomes invalid (⇒ bad file descriptor).
 - The inotify fd is recycled by the OS (⇒ undefined bahavior).

I'd be happy to submit a PR to fix this. However, any solution I can come up with involves at least some API breakage. Therefore I'd really like to get some input first.

(ref passcod/rsnotify#37)

hannobraun:
Thanks for reporting!

I think the root issue is that `INotify` allows the inotify API to be misused in such a way. I wrote it when I had much less experience with Rust, and it clearly shows.

> I'd be happy to submit a PR to fix this. However, any solution I can come up with involves at least some API breakage.

I'd be very thankful for that! Any amount of breakage is fine, as the API is so clearly wrong :)

> Therefore I'd really like to get some input first.

What I would do first, is read the inotify documentation carefully, to find out what the story about multi-threaded usage is. If it's allowed, there should be a way to safely expose it. If not, the wrapper needs to reflect that.

Both outcomes would be fine with me. The priority is to provide a correct API. If it comes down to restricting the API, it would make sense to coordinate with rsnotify, to make sure their needs are met.

PR #29, opened 2015-10-30
Allow asynchronous closing.

benschulz:
Alright, I'm by no means done, but I thought I'd give you a chance to weigh in earlier, rather than later.

**Changes**
 - `INotify` is no longer cloneable. Therefore the API is essentially single-threaded, enforced by ownership semantics.
 - A new method `closer()` is introduced. Its return value enables asynchronous closing of the inotify fd.
   - To avoid any races, the actual closing is still done by whoever owns the corresponding `INotify`.
   - This only happens when `wait_for_events()` is called.
   - To allow `wait_for_events()` to react to asynchronous closing, it spins in an epoll loop with a smallish timeout.

**TODO**
 - [ ] Look at `available_events()` wrt asynchronous closing. Currently it does not make sense, because the caller can't determine if the inotify fd was closed without calling a blocking method. Maybe expose the state?
 - [ ] Figure out a good epoll timeout (10ms seems a bit high).
 - [x] Go over all the `unwrap()` calls and figure out how to deal with the potential error conditions.
 - [ ] More tests
 - [ ] Documentation

(ref #28)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,3 +16,7 @@ license = "ISC"
 
 [dependencies]
 libc = "*"
+epoll = "*"
+
+[dev-dependencies]
+tempdir = "*"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,6 +19,7 @@
 //! [wiki]: https://en.wikipedia.org/wiki/Inotify
 //! [inotify7]: http://man7.org/linux/man-pages/man7/inotify.7.html
 
+extern crate epoll;
 extern crate libc;
 
 pub use wrapper::INotify;
--- a/src/wrapper.rs
+++ b/src/wrapper.rs
@@ -314,7 +314,7 @@ impl Drop for INotify {
 
 impl INotifyCloser {
     pub fn close_async(&self) {
-        if state == INotifyState::Open {
+        if self.state() == INotifyState::Open {
             self.state.store(CLOSING, Ordering::Relaxed);
         }
     }
--- a/src/wrapper.rs
+++ b/src/wrapper.rs
@@ -5,6 +5,7 @@
 use epoll::{self, EpollEvent};
 use epoll::util::ctl_op as EpollControlOp;
 use epoll::util::event_type as EpollEventType;
+use epoll::util::WaitError;
 use libc::{
     F_GETFL,
     F_SETFL,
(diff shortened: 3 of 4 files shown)


hannobraun:
I've only had time to take a superficial look so far, so please disregard the following comments, if they don't make sense:
- Do you think it would be practical to move all that new complexity to a wrapper around `INotify`, to keep `INotify` itself simple and single-threaded?
- Would it be possible to revert `fd` back to `c_int` (from `Option<c_int>`) and consume the `INotify` instance in `close` (by changing the self argument from `&mut self` to `self`) to prevent misuse?

Thank you for all your work so far! I'll take a closer look within the next few days.

benschulz:
> Would it be possible to revert fd back to c_int (from Option<c_int>) and consume the INotify instance in close (by changing the self argument from &mut self to self) to prevent misuse?

That does not account for asynchronous closing. I guess the `closer` could "compensate" for that. So yeah.

> Do you think it would be practical to move all that new complexity to a wrapper around INotify, to keep INotify itself simple and single-threaded?

I've thought about this for a bit. There are three possible levels of abstraction.

 - Raw FFI
 - A super-thin layer atop the FFI, where e.g. functions return `Result`s and error codes are mapped to enum values. It gives access to the `RawFd` so one can work with `select(2)`/epoll and write one's own event loop.
 - A higher level abstraction which hides the `RawFd` and a lot of the other internals. Convenient, but less powerful; one can't write an event loop which waits for IO readiness of the inotify fd and other resources.

The third layer comes with a decent amount of complexity and there is no way around that. I'm wondering what inotify-rs is supposed to provide. I get the feeling your original use case requires the second, more powerful layer. If that's the case then perhaps we should just make the original API single-threaded (i.e. no cloning) and the third layer has to be implemented downstream…

hannobraun:
> I'm wondering what inotify-rs is supposed to provide.

That's the key question here. I originally developed the library for my own use in an application that I'm no longer working on. The current `INotify` wrapper grew out of that, and the pull requests that have trickled in since then.

> I get the feeling your original use case requires the second, more powerful layer. If that's the case then perhaps we should just make the original API single-threaded (i.e. no cloning) and the third layer has to be implemented downstream…

The more I think about it, the more that seems like the right solution to me. According to the download stats on crates.io[[1]](https://crates.io/crates/inotify)[[2]](https://crates.io/crates/notify), it seems that a significant number of inotify-rs users depend on it indirectly via rsnotify. But I'd say there must be another significant faction of users that use it directly or via other libraries that I don't know about (inotify-rs has a lot more downloads overall, and has download spikes on days when rsnotify has very few downloads).

Before yours, I haven't received any issue reports for months, so I assume most users are reasonably happy. The problems that you reported would be solved by tightening the API (no cloning, as you suggest; `close` consuming the `INotify` instance) and implementing the higher-level wrapper in fsnotify.

That seems like the best course of action to me. If it turns out other inotify-rs users were relying on (the broken) multithreading support, we can move the wrapper upstream again. What do you think?

benschulz:
> What do you think?

I agree. I opened another pull request to remove the `#[derive(Clone)]`.

hannobraun:
Thanks a lot for your work, @benschulz. I hope you can use most of what you wrote for this pull request downstream in rsnotify without too much trouble.

Sorry that I wasn't able to give you better guidance from the beginning.

benschulz:
With open source development becoming mainstream, I'm generally happy to get a replay *at all*. Yours were polite and well reasoned on top of that. In other words: Please don't apologize. ;)

hannobraun:
:)

PR #30, opened 2015-11-03
INotify is no longer cloneable.

benschulz:
The ability to clone `INotify` enabled races (see #28 and #29).


hannobraun:
Thanks!

PR #31, opened 2015-11-03
Fixed sporadic test failure.

benschulz:
Cargo runs tests in parallel. A dev-dependency on [tempdir][1] was added to cleanly separate tests. This fixes the sporadic failure of `it_should_not_return_duplicate_events` when other tests interfered with it.

[1]: https://github.com/rust-lang-nursery/tempdir


hannobraun:
Looks good, thank you!

PR #32, opened 2016-01-28
Update to libc 0.2

AerialX:
See https://github.com/rust-lang/rust/issues/29867


hannobraun:
Thanks a lot!

The Travis failure seems to be caused by our out-of-date `.travis.yml`. I've tested locally on stable and nightly and had no issue there. Merging.

PR #54, opened 2017-05-28
Update documentation for read_events{,_blocking}

tokenrove:
The names of these methods were changed in commit
  de56c15c5cf0d95f900aed9c6bb789158193fc38
but the documentation still referred to the old methods.


hannobraun:
Thank you!

PR #55, opened 2017-07-25
deriving traits for WatchDescriptor

arsalan86:
I've derived the Hash trait for WatchDescriptor, so it can be added to HashMaps and used for lookup.

I ran into difficulties while implementing inotify-rs in a project that watches a large number of files. Not being able to match an inotify::Event to a file in a database is crippling for me.


inotify/src/lib.rs
```diff
@@ -33,6 +33,7 @@ use std::os::unix::ffi::OsStrExt;
 use std::os::unix::io::RawFd;
 use std::path::Path;
 use std::slice;
+use std::collections::HashMap;
```
shepmaster: Do you really need to include this? I'd think you only need to add the `derive(Hash)` below.

shepmaster:
I'm not a maintainer, but I'd bet you could squash all of those commits, especially since they only provide a single line of change.

shepmaster:
I think that the type could also implement the three `RawFd` traits in https://doc.rust-lang.org/std/os/unix/io/index.html

arsalan86:
commits squashed.

Instead of deriving Hash for the WatchDescriptor struct, I've made the field RawFd public. This works for me, and is much cleaner than using a struct as a HashMap key.

inotify/src/lib.rs
```diff
@@ -27,10 +27,750 @@ extern crate libc;
...
+///
+/// Can be obtained from [`Inotify::add_watch`] or from an [`Event`]. A watch
+/// descriptor can be used to get inotify to stop watching a file by passing it
+/// to [`Inotify::rm_watch`].
+///
+/// [`Inotify::add_watch`]: struct.Inotify.html#method.add_watch
+/// [`Inotify::rm_watch`]: struct.Inotify.html#method.rm_watch
+/// [`Event`]: struct.Event.html
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct WatchDescriptor(pub RawFd);
```
arsalan86: changed field RawFd to public to allow consumers to read it

shepmaster:
> commits squashed.

I think you might have had some crossed wires in your rebase; the commit is now +945 −789. 

Again, I'm not a maintainer so you don't need to listen to me, but were this my project I'd be very hesitant about exposing a field. The traits from the standard library would be more comfortable to me (and then you also get to participate in that abstraction).

> is much cleaner than using a struct as a `HashMap` key.

I don't necessarily agree, but I can't speak to your particular usecase.

inotify/src/lib.rs
arsalan86: The Hash trait is part of std::collections::HashMap

arsalan86:
Yes, I screwed that rebase up pretty bad. I should brush up on git and best practices.

I understand the hesitation re: exposing a field.

I'm going to try and figure out if I can implement the traits.

arsalan86:
I'm going to close this PR until I've arrived at a working solution that would benefit other users of inotify-rs.

hannobraun:
Thank you @arsalan86 for the pull request, and thank you @shepmaster for the review!

I've [opened an issue](https://github.com/hannobraun/inotify-rs/issues/58) regarding the `RawFd` traits. Thanks for the suggestion!

PR #57, opened 2017-07-26
add: Hash trait for WatchDescriptor

arsalan86:
I've made a new PR after breaking the previous one due to a clunky rebase/squash.

I've added the `Hash` to the `WatchDescriptor` struct, to allow it to be used for `HashMap` lookup.

### DIFF
--- a/inotify/src/lib.rs
+++ b/inotify/src/lib.rs
@@ -557,7 +557,7 @@ pub use self::watch_mask::WatchMask;
 /// [`Inotify::add_watch`]: struct.Inotify.html#method.add_watch
 /// [`Inotify::rm_watch`]: struct.Inotify.html#method.rm_watch
 /// [`Event`]: struct.Event.html
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
 pub struct WatchDescriptor(RawFd);
 
 


inotify/src/lib.rs
```diff
@@ -33,6 +33,7 @@ use std::os::unix::ffi::OsStrExt;
 use std::os::unix::io::RawFd;
 use std::path::Path;
 use std::slice;
+use std::collections::HashMap;
```
hannobraun: You should be able to remove this. Deriving traits shouldn't require any imports.

The compiler also [warns about this import being unused](https://travis-ci.org/hannobraun/inotify-rs/jobs/257566812#L260).

arsalan86:
I've implemented the requested changes, thanks!

hannobraun:
Can you squash the commits into a single one? That way it's easier to read later. If you're unsure about how to do that, something like `git rebase -i origin/master` should get you on the way.

Again, if you don't have time or can't be bothered, let me know. I can do it later, no problem. Thanks!

arsalan86:
Done!

hannobraun:
Hmm, still seeing two commits. Maybe you forgot to push the branch?

arsalan86:
It's at 4 commits now, I think I'm doing something wrong! I'm sorry!

hannobraun:
No worries! Git can be quite complicated if you're not used to it. That third commit is really weird. It just repeats the change from the third commit, which shouldn't really be possible. Oh well :-)

I'm happy to take care of this for you, but if you want to figure out how to solve this yourself, here's what I would try:
1. First I would make a second branch to experiment on, so you don't accidentally lose the commits. Like this: `git checkout -b backup`, where `backup` can be any valid branch name. That command creates the branch `backup` and switches over to it.
2. Then I would get rid of the last two commits, like this: `git reset --hard HEAD~2`. This resets the branch to your second commit, essentially removing the last two commits from the branch. This can be quite dangerous, that's why we made the backup branch earlier. See [1] for more explanation of `HEAD~2`.
3. You should be left with the first two commits now. Run `git rebase -i origin/master` or `git rebase -i HEAD~2`. Unless I've made a mistake, both `origin/master` and `HEAD~2` should point to the same commit at this point. This should open a text editor which should show your last two commits.
4. In the text editor for the interactive rebase, locate the line of your second commit. In that line, replace `pick` with `fixup`. Then save and exit the editor. If everything goes right, Git should now merge the second commit into the first one.
5. You should be left with a single commit. You can push that to your repository now. However, since that push would change an existing commit and delete others, you have to add the `-f` option to the `git push` command.

As I said, I'm happy to do it for you. I just figured you might want some pointer on how to solve such problems with Git in the future. Let me know if you want me to do it, or if you have any questions!

[1] `HEAD` always refers to the last commit of the current branch. `HEAD~1` is the commit before that, and `HEAD~2` is the commit before that one. So by resetting to `HEAD~2`, the last two commits are removed from the branch.

arsalan86:
Done!

Thanks for the git help!

I'm quite new to git/github, and still figuring things out. I will take some time out to go through the man pages.

Thank you for your patience.

hannobraun:
You're welcome, and thanks again for the pull request!

And regarding Git, it's actually quite elegant under the surface. The only problem is that the elegance is hidden under a thick layer of weirdness :-)

Issue #59, opened 2017-07-26
WatchDescriptors returned by Inotify are arbitrary integers, not Unix File Descriptors


arsalan86:
The inotify-rs wrapper makes use of the `std::os::unix::io::RawFd` type to describe Inotify `WatchDescriptor`s.

Inotify itself describes WatchDescriptors as an integer:

    struct inotify_event {
            int      wd;       /* Watch descriptor */
            uint32_t mask;     /* Mask describing event */
            uint32_t cookie;   /* Unique cookie associating related
                                    events (for rename(2)) */
            uint32_t len;      /* Size of name field */
            char     name[];   /* Optional null-terminated name */
        };

sourced from: http://man7.org/linux/man-pages/man7/inotify.7.html

A closer examination of the Inotify source reveals that WatchDescriptors are, in fact, arbitrary integers used to identify Events with the calls to add_watch() that generated them, as outlined [in this stackoverflow answer](https://stackoverflow.com/a/45316225/3818171)

The inotify-rs wrapper should not use the `RawFD` type to refer to WatchDescriptors, and should use an `i32` instead.

hannobraun:
Oh, good catch! Although I don't thin `i32` is right, as the size of C's `int` can vary depending on platform/compiler. `libc::c_int` should be correct, though.

Thanks for pointing out that Stack Oveflow answer. Some good points in there. I'm going to open issues for the problems mentioned there.

arsalan86:
I'm working on this issue, seems like an easy fix.

Do you think it's reasonable to add a method to the `WatchDescriptor` struct that returns the contained field?

This allows for easier handling of WatchDescriptors, at least in my case.

hannobraun:
> Do you think it's reasonable to add a method to the WatchDescriptor struct that returns the contained field?

Yes, totally. Feel free to send in a pull request.

Thanks!

PR #63, opened 2017-07-26
bugfix: inotify-rs issue #59: using libc::c_int instead of RawFd for …

arsalan86:
…WatchDescriptors

Fixes Issue #59 

changed WatchDescriptor(RawFd) to WatchDescriptor(c_int) and added import for clib::c_int


hannobraun:
Thanks!

Issue #60, opened 2017-07-26
`Eq` implementation of `WatchDescriptor` is wrong


hannobraun:
From [Stack Overflow](https://stackoverflow.com/questions/45315521/how-do-i-use-the-watchdescriptor-struct-generated-by-inotify-rs-to-match-events/45316225#45316225):

> As noted above, inotify watch descriptor values are scoped per parent inotify file descriptor. This makes the use of [derive(Eq)] on WatchDescriptor highly questionable. Looks like a straight-up bug.

I haven't had the chance to look into the details myself so far, but this seems legit.

arsalan86:
I have tested this by creating multiple `Inotify` instances and matching the `WatchDescriptors` they generate.

`assert_eq!(wd_from_inotify1, wd_from_inotify2);` should not be possible.

While the `Eq` trait does seem useful within the scope of an `Inotify` instance, the potential to match across instances can break things.

I don't think removing the `Eq` trait is the right way to address this issue. Perhaps it makes more sense for the `Inotify` struct to generate an identifier unique to itself. This can be passed to generated `WatchDescriptor`s as a field. This would retain the usability of the `Eq` trait within the scope of the parent `Inotify`, and protect against matching `WatchDescriptor`s between instances of `Inotify`

hannobraun:
I think the most straight-forward way to address this would be to add the inotify instance's file descriptor to `WatchDescriptor`. That would do nothing to solve #61, but at least the `Eq` implementation would be working correctly.

arsalan86:
Yes, that does seem to be the most practical way of fixing the `Eq` implementation. I've implemented a field of type `RawFd` in the `WatchDescriptor` struct in my working tree.

I'll do a PR once I've figured out what can be done about #61. My tree doesn't build at the moment.

hannobraun:
If you make a pull request for adding the `RawFd', I'd definitely accept it. Yes, eventually we want a complete solution that also solves #61, but I suspect that will require more thinking and therefore more time. Until then, fixing the `Eq` issue in a simple way will be very welcome.

arsalan86:
Adding the `RawFd` field to `WatchDescriptor` introduces another issue: Inotify itself doesn't associate watch descriptors and events with a particular Inotify instance.

From the Inotify source (accessed from [here](http://elixir.free-electrons.com/linux/latest/source/fs/notify/inotify/inotify_user.c#L189)):

	pad_name_len = round_event_name_len(fsn_event);
[...]

PR #67, opened 2017-08-07
bugfix: issue #60: added RawFd field to WatchDescriptor, updated func…

arsalan86:
…tions.

### DIFF
--- a/inotify/src/lib.rs
+++ b/inotify/src/lib.rs
@@ -29,6 +29,7 @@ extern crate inotify_sys as ffi;
 
 use std::mem;
 use std::io;
+use std::io::ErrorKind;
 use std::os::unix::ffi::OsStrExt;
 use std::os::unix::io::RawFd;
 use std::path::Path;
@@ -211,7 +212,7 @@ impl Inotify {
 
         match wd {
             -1 => Err(io::Error::last_os_error()),
-            _  => Ok(WatchDescriptor(wd)),
+            _  => Ok(WatchDescriptor{id: wd, fd: self.0}),
         }
     }
 
@@ -264,7 +265,10 @@ impl Inotify {
     /// [`Inotify::add_watch`]: struct.Inotify.html#method.add_watch
     /// [`Event`]: struct.Event.html
     pub fn rm_watch(&mut self, wd: WatchDescriptor) -> io::Result<()> {
-        let result = unsafe { ffi::inotify_rm_watch(self.0, wd.0) };
+        if self.0 != wd.fd {
+            return Err(io::Error::new(ErrorKind::InvalidInput, "Invalid WatchDescriptor"))
+        }
+        let result = unsafe { ffi::inotify_rm_watch(self.0, wd.id) }; 
         match result {
             0  => Ok(()),
             -1 => Err(io::Error::last_os_error()),
@@ -366,7 +370,7 @@ impl Inotify {
             -1 => {
                 let error = io::Error::last_os_error();
                 if error.kind() == io::ErrorKind::WouldBlock {
-                    return Ok(Events::new(buffer, 0));
+                    return Ok(Events::new(self.0, buffer, 0));
                 }
                 else {
                     return Err(error);
@@ -396,7 +400,7 @@ impl Inotify {
             }
         };
 
-        Ok(Events::new(buffer, num_bytes))
+        Ok(Events::new(self.0, buffer, num_bytes))
     }
 
     /// Closes the inotify instance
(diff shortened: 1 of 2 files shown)


hannobraun:
Thanks a lot! Sorry for my late reply here, I'm a bit preoccupied at the moment. I'll take a closer look as soon as I can. It's possible that I won't get a chance until next week though.

inotify/src/lib.rs
```diff
@@ -264,10 +265,17 @@ impl Inotify {
     /// [`Inotify::add_watch`]: struct.Inotify.html#method.add_watch
     /// [`Event`]: struct.Event.html
     pub fn rm_watch(&mut self, wd: WatchDescriptor) -> io::Result<()> {
-        let result = unsafe { ffi::inotify_rm_watch(self.0, wd.0) };
+        let result: c_int;
+        if self.0 == wd.1 {
+            result = unsafe { ffi::inotify_rm_watch(self.0, wd.0) }; 
+        }
+        else {
+            result = -2;
```
hannobraun: I don't like how you use `result` here, for the following reasons:
1. Using an integer variable to express the error condition is not type-safe. The only reason we have that here at all, is because of the underlying inotify API.
2. You're overloading the meaning of `result`. Before it was about one thing only (the return value from `inotify_rm_watch`), now there's a second concern (whether the `WatchDescriptor` is valid) mixed into it.

To be clear, I believe this works (haven't tested it), but the code is not as clear as it could be. I'd just leave the two things separate: Check the watch descriptor at the beginning and do an early return if it's invalid. Leave the rest of the method as it is.

inotify/src/lib.rs
```diff
@@ -559,7 +567,7 @@ pub use self::watch_mask::WatchMask;
 /// [`Inotify::rm_watch`]: struct.Inotify.html#method.rm_watch
 /// [`Event`]: struct.Event.html
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
-pub struct WatchDescriptor(c_int);
+pub struct WatchDescriptor(c_int, RawFd);
```
hannobraun: I think it makes sense to turn `WatchDescriptor` into a struct with named fields at this point, but this works too.

inotify/src/lib.rs
arsalan86: Thanks for the feedback!

Yes, you're right, it does make sense to not pollute `result`. I'll fix this soon.

inotify/src/lib.rs
arsalan86: I'll go ahead and name the fields.

arsalan86:
Re: above changes. Things are a bit hectic for me this week, but I should be able to get the changes in this coming weekend.

hannobraun:
Thanks, sounds great. Take your time!

arsalan86:
I've made the requested changes.

inotify/tests/main.rs
```diff
@@ -69,6 +69,20 @@ fn it_should_handle_file_names_correctly() {
...
+    let mut testdir = TestDir::new();
+    let (path, file) = testdir.new_file();
+
+    let mut inotify = Inotify::init().unwrap();
+    let wd1 = inotify.add_watch(&path, watch_mask::ACCESS).unwrap();
+
+    let mut second_inotify = Inotify::init().unwrap();
+    let wd2 = second_inotify.add_watch(&path, watch_mask::ACCESS).unwrap();
+
+    assert!(wd1 != wd2);
```
hannobraun: I'd like to see another assert to test the new if condition in `rm_watch`. Something like this:
```
assert_eq!(inotify.rm_watch(wd2).unwrap_err().kind, ErrorKind::InvalidInput);
```

arsalan86: got it, gimme a couple of minutes to commit this

arsalan86: good call! I am not able to get it to behave in that manner. Behavior suggests file descriptors returned by inotify_init1 are either the same for multiple instances, or they test true for == comparisons.

hannobraun: That's very weird. Let me know if you need help figuring this out. I can take a look later.

arsalan86: Found the problem, fixed it. Silly mistake on my part, using a '==' in place of a '!='.

hannobraun:
Thanks again, this is good to go now. I'm not going to merge it right away, as I want to release version 0.4.2 first. As this is a breaking change, it will require a 0.5 version.

I plan to make the relase next week at the latest, and will merge this right after.

hannobraun:
0.4.2 ist out. Merging.

PR #98, opened 2018-02-13
typos: fix typos in comments

mathstuf:


mathstuf:
Not sure what's up with the test failure; I dodn't change the doctest which is failing and all the other changes are in comments.

hannobraun:
Thanks a lot!

That test failure is unrelated. I actually have a commit in a local branch that removes the failing example, but somehow forgot all about it until looking into this right now. I'll take care of that later. Thanks again!

PR #99, opened 2018-02-13
feat: support async polling via futures

### LINKED ISSUE #49: Base API on futures-rs
The current API offers both synchronous and asynchronous operation, but it would be better to do that based on [futures-rs](https://github.com/alexcrichton/futures-rs). I've looked into it, but decided to hold off for the moment, until I had a chance to learn a bit more about futures-rs and experience it from a user's perspective (as opposed to that of a library author).

mathstuf:
Closes #49.

---
No idea how idiomatic it is, but the test works :) .

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -74,6 +74,7 @@
 #[macro_use]
 extern crate bitflags;
 
+extern crate futures;
 extern crate libc;
 extern crate inotify_sys as ffi;
 
@@ -100,10 +101,14 @@ use std::sync::{
 use std::slice;
 use std::ffi::{
     OsStr,
+    OsString,
     CString,
 };
 
+use futures::{Async, Poll, Stream};
+use futures::task;
 use libc::{
+    FILENAME_MAX,
     F_GETFL,
     F_SETFL,
     O_NONBLOCK,
@@ -465,6 +470,19 @@ impl Inotify {
         Ok(Events::new(Arc::downgrade(&self.fd), buffer, num_bytes))
     }
 
+    /// Create a stream which collects events
+    ///
+    /// Returns a `Stream` over all events that are available. This stream is an
+    /// infinite source of events.
+    ///
+    /// Note that this stream is not optimal and always reschedules itself if a
+    /// read would block.
+    ///
+    /// An internal buffer which can hold the maximum possible size is used.
+    pub fn event_stream(&mut self) -> EventStream {
+        EventStream::new(self.fd.clone())
+    }
+
     /// Closes the inotify instance
     ///
     /// Closes the file descriptor referring to the inotify instance. The user
@@ -543,6 +561,95 @@ fn read_into_buffer(fd: RawFd, buffer: &mut [u8]) -> isize {
     }
 }
 
+// Use this when 1.24 is available for use. We can use the hard-coded 16 due to Linux's ABI
+// guarantees.
+// const EVENT_MAX_SIZE: usize = mem::size_of::<ffi::inotify_event>() + (FILENAME_MAX as usize) + 1;
+const EVENT_MAX_SIZE: usize = 16 + (FILENAME_MAX as usize) + 1;
+
+/// Stream of inotify events
+///
+/// Allows for streaming events returned by [`Inotify::event_stream`].
+///
+/// [`Inotify::event_stream`]: struct.Inotify.html#method.event_stream
+pub struct EventStream {
+    fd: Arc<RawFd>,
+    buffer: [u8; EVENT_MAX_SIZE],
+    pos: usize,
+    size: usize,
+}
+
+impl EventStream {
+    fn new(fd: Arc<RawFd>) -> Self {
+        EventStream {
+            fd: fd,
+            buffer: [0; EVENT_MAX_SIZE],
+            pos: 0,
+            size: 0,
+        }
+    }
+}
+
+impl Stream for EventStream {
+    type Item = EventOwned;
+    type Error = io::Error;
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error>
+    {
+        if 0 < self.size {
+            let (step, event) = Event::from_buffer(Arc::downgrade(&self.fd), &self.buffer[self.pos..], self.size);
+            self.pos += step;
+            self.size -= step;
+
+            return Ok(Async::Ready(Some(event.into_owned())));
+        }
+
+        let num_bytes = read_into_buffer(*self.fd, &mut self.buffer);
+
+        let num_bytes = match num_bytes {
+            0 => {
+                return Ok(Async::Ready(None));
+            }
+            -1 => {
+                let error = io::Error::last_os_error();
+                if error.kind() == io::ErrorKind::WouldBlock {
+                    task::current().notify();
+                    return Ok(Async::NotReady);
+                }
+                else {
+                    return Err(error);
+                }
+            },
+            _ if num_bytes < 0 => {
+                panic!("{} {} {} {} {} {}",
+                    "Unexpected return value from `read`. Received a negative",
+                    "value that was not `-1`. According to the `read` man page",
+                    "this shouldn't happen, as either `-1` is returned on",
+                    "error, `0` on end-of-file, or a positive value for the",
+                    "number of bytes read. Returned value:",
+                    num_bytes,
+                );
+            }
+            _ => {
+                // The value returned by `read` should be `isize`. Let's quickly
+                // verify this with the following assignment, so we can be sure
+                // our cast below is valid.
+                let num_bytes: isize = num_bytes;
+
+                // The type returned by `read` is `isize`, and we've ruled out
+                // all negative values with the match arms above. This means we
+                // can safely cast to `usize`.
+                debug_assert!(num_bytes > 0);
+                num_bytes as usize
+            }
+        };
+
+        let (step, event) = Event::from_buffer(Arc::downgrade(&self.fd), &self.buffer, num_bytes);
+        self.pos = step;
+        self.size = num_bytes - step;
+
+        Ok(Async::Ready(Some(event.into_owned())))
+    }
+}
 
 bitflags! {
     /// Describes a file system watch
@@ -872,7 +979,7 @@ impl<'a> Events<'a> {
 }
 
 impl<'a> Iterator for Events<'a> {
-    type Item = Event<'a>;
+    type Item = Event<&'a OsStr>;
 
     fn next(&mut self) -> Option<Self::Item> {
         if self.pos < self.num_bytes {
@@ -899,7 +1006,7 @@ impl<'a> Iterator for Events<'a> {
 /// [`Inotify::read_events_blocking`]: struct.Inotify.html#method.read_events_blocking
 /// [`Inotify::read_events`]: struct.Inotify.html#method.read_events
 #[derive(Clone, Debug)]
-pub struct Event<'a> {
+pub struct Event<S> {
     /// Identifies the watch this event originates from
     ///
     /// This [`WatchDescriptor`] is equal to the one that [`Inotify::add_watch`]
@@ -931,10 +1038,10 @@ pub struct Event<'a> {
     /// This field is set only if the subject of the event is a file in a
     /// watched directory. If the event concerns a file or directory that is
     /// watched directly, `name` will be `None`.
-    pub name: Option<&'a OsStr>,
+    pub name: Option<S>,
 }
 
-impl<'a> Event<'a> {
+impl<'a> Event<&'a OsStr> {
     fn new(fd: Weak<RawFd>, event: &ffi::inotify_event, name: &'a OsStr)
         -> Self
     {
@@ -1036,7 +1143,19 @@ impl<'a> Event<'a> {
 
         (step, event)
     }
+
+    fn into_owned(&self) -> EventOwned {
+        Event {
+            wd: self.wd.clone(),
+            mask: self.mask,
+            cookie: self.cookie,
+            name: self.name.map(OsStr::to_os_string),
+        }
+    }
 }
+
+/// An owned version of `Event`
+pub type EventOwned = Event<OsString>;
 
 
 bitflags! {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ travis-ci   = { repository = "inotify-rs/inotify" }
 
 [dependencies]
 bitflags    = "1"
+futures     = "0.1"
 inotify-sys = "0.1.1"
 libc        = "0.2"
 
--- a/tests/main.rs
+++ b/tests/main.rs
@@ -4,6 +4,7 @@
 // This test suite is incomplete and doesn't cover all available functionality.
 // Contributions to improve test coverage would be highly appreciated!
 
+extern crate futures;
 extern crate inotify;
 extern crate tempdir;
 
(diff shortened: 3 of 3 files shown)


mathstuf:
Ideally, the common code between the implementations would be shared. The biggest difference is the `Rc`/`Arc` stuff and `async::Event` stores a buffer rather than using an externally-provided one.

mathstuf:
Added a WIP commit which makes the core code `Send`-able which allows for unification of most of the types.

hannobraun:
Thank you for this pull request, @mathstuf!

I haven't had time to look into the details yet. I've added it to my task list and I hope I can get around to it by next week, at the latest.

Some preliminary comments:
- Duplicating all the code for async is highly undesirable, so sharing as much as possible seems like the way to go. If all it takes is replacing some `Rc`s with `Arc`, then I have no problem with that.
- Possibly stupid question (as I said, haven't had time to look into the details yet): Do you think it's practical to completely replace `read_events`/`read_events_blocking` and just use [`Stream::wait`](https://docs.rs/futures/0.1.18/futures/stream/trait.Stream.html#method.wait) for the synchronous use case? That way, we wouldn't have to support two different APIs.

mathstuf:
> Duplicating all the code for async is highly undesirable, so sharing as much as possible seems like the way to go. If all it takes is replacing some Rcs with Arc, then I have no problem with that.

OK, I'll squash the commit then.

> Possibly stupid question (as I said, haven't had time to look into the details yet): Do you think it's practical to completely replace read_events/read_events_blocking and just use Stream::wait for the synchronous use case? That way, we wouldn't have to support two different APIs.

It'd be possible if `Event` became owning rather than borrowing from a buffer. It also means that the buffer size is the maximum possible size (unless a `Vec<u8>` is acceptable) which is ~4k.

As for the actual implementation, it'd be hard; the `Stream` isn't going to know when it is "done" to return a set of events. There may be a possibility with something like this code (mocked up):

```rust
lazy_static! {
    static ref TIMER_WHEEL: Timer = Timer::default();  // From the `tokio-timer` crate.
    static ref TIME_LIMIT: Duration = Duration::from_millis(10);
}

fn read_events(&self) -> Result<Vec<Event>> {
    TIMER_WHEEL.timeout_stream(self.by_ref(), *TIME_LIMIT).collect().wait()
}

fn read_events_blocking(&self) -> Result<Event> {
    self.by_ref().take(1).wait()
}
```

Note that the [`tokio-timer`](https://github.com/tokio-rs/tokio-timer) crate is just maintained by Tokio developers; it is usable from `futures` directly.

hannobraun:
@mathstuf Thanks for the info. Unless required for safety reasons, I'd like to avoid adding much overhead on top of inotify, which means stuff like `Vec` is only acceptable if there's a really good reasons for it.

I'll think more about it later, when I had a chance to take a closer look.

mathstuf:
OK, I guess there might be an argument to have a crate-level feature to increase the buffer size of `Event`. Something like:

```rust
// feature(full)
EVENT_MAX_SIZE = 8209
// feature(reasonable)
EVENT_MAX_SIZE = 512
// default
EVENT_MAX_SIZE = 128
```

And applications can choose the level that they want, though insufficient sizes have interesting behaviors according to `inotify(7)`:

    in kernels before 2.6.21, read(2) returns 0; since kernel 2.6.21, read(2) fails with the error EINVAL.

mathstuf:
Oh, nevermind, I have it currently storing in `OsString`, so `Event` isn't excessive, but it is using the heap.

src/lib.rs
```diff
@@ -1024,6 +926,79 @@ impl<'a> Event<'a> {
             name  : name,
         }
     }
+
+    fn from_buffer(fd: Weak<RawFd>, buffer: &[u8], num_bytes: usize) -> Self {
```
mathstuf: This currently extracts a single event per `read` call. There will need to be some buffer management in order to extract multiple events from a single `read`.

hannobraun:
@mathstuf I've taken another look and I'm not too happy about all those additional heap allocations when reading events. I really like that the current API follows inotify in that it lets the user pass the buffer. If at all feasible, I'd like to preserve that.

Are you still experimenting or do you want me to do a full review? Anything else I can do to help?

mathstuf:
On February 16, 2018 9:18:41 AM UTC, Hanno Braun <notifications@github.com> wrote:
>@mathstuf I've taken another look and I'm not too happy about all those
>additional heap allocations when reading events. I really like that the
>current API follows inotify in that it lets the user pass the buffer.
>If at all feasible, I'd like to preserve that.

The problem is that the stream would have to stop and wait (Ok(Async::NotReady)) for consumption of all events in the current buffer before it can read again (or at least all events in the first MAX_EVENT_SIZE bytes). And I have no idea how to do such a thing in Rust off hand. The lifetimes just get unwieldy. I guess that some Arc/Weak magic could be done, but I don't want to think about that this early in the morning :) .

>Are you still experimenting or do you want me to do a full review?
>Anything else I can do to help?

I won't be able to work on it until Monday, so feel free to review. The commits are split just to help with tracking down regressions; I'll squash it all down before merging.

One thing that could probably be done better is to use a better Read wrapper around the fd (e.g., from Tokio) and doing and_then off of that stream. But the buffer management gets nasty again, but it could probably be done better by making a stream of Vec<Event> and flatten(). Currently, print statements shows that this is basically a busy loop when the loop is waiting for data.

hannobraun:
I didn't do a line-by-line review, but I've taken another look at the changes. I have some questions and comments:
- I've decided that stuff like returning `Vec` is not acceptable for a low-level library like this. I'd like to provide a `futures`-based API without adding this additional overhead. If that's not possible, it's probably best to provide that in a separate wrapper library (like [tokio-inotify](https://crates.io/crates/tokio-inotify), although that's not up-to-date).
- Is doing the reading manually the best approach here? There's stuff like [PollEvented](https://docs.rs/tokio/0.1.1/tokio/reactor/struct.PollEvented.html) and [EventedFd](https://docs.rs/mio/0.6.13/mio/unix/struct.EventedFd.html) which look like they solve this problem, but I haven't had a chance to look into it more deeply.
- If the goal is to do this without managing buffers, then maybe implementing `Stream` is the wrong approach. A single read call can return multiple events, so maybe the existing `read_events` function (that takes a buffer) could be modified to return a `Stream<Event>`, or a `Future<Iterator<Event>>` (those types are only meant to demonstrate my point, and shouldn't be taken literally) `EventStream` may be a step in the right direction. Could it be changed to borrow `buffer` instead of owning it?

Sorry that I couldn't be more helpful so far. I haven't had that much time to look into this, and my lack of knowledge about futures/tokio makes it hard to assess any of this.

mathstuf:
> I've decided that stuff like returning Vec is not acceptable for a low-level library like this. I'd like to provide a futures-based API without adding this additional overhead.

I can make the old methods continue to return `Iterator` structs. The `OsString` is harder to avoid though (see below). An alternative is to make `Event<'a>` used by the sync methods and `EventOwned` (or some other name) used by the async method.

> If that's not possible, it's probably best to provide that in a separate wrapper library (like tokio-inotify, although that's not up-to-date).

Unless `Event::from_buffer` is getting exposed somehow, there's little in reusing the logic here; it'd be better just to reimplement.

> Is doing the reading manually the best approach here?

I mentioned this is my previous comment. Wrapping is probably better. I haven't looked at which is best though.

> If the goal is to do this without managing buffers, then maybe implementing Stream is the wrong approach. A single read call can return multiple events, so maybe the existing read_events function (that takes a buffer) could be modified to return a Stream<Event>, or a Future<Iterator<Event>> (those types are only meant to demonstrate my point, and shouldn't be taken literally) EventStream may be a step in the right direction.

Without a `Stream`, there isn't much use for `futures`. If I have to keep going back to the origin and pump the `read` call again manually, I'd be disinterested in how `futures` is supported in this crate.

> Could it be changed to borrow buffer instead of owning it?

As I said before, the lifetime tracking is heinous. The `Stream` wouldn't be allowed to read more than the buffer size at one time ever and it'd have to track the life of all generated events. Basically, trying to store an `Event` would stall the stream. That sounds like a nasty deadlock sitting in everyone's path.

mathstuf:
Updated to keep a buffer for the `sync` side at least. I've shared as much code as possible (the only bit which is duplicated is docs for `Event` versus `EventOwned` and the handling of return values from the `read` call. Looking into how to use fancy `tokio-io` bits now.

mathstuf:
I added a commit which uses the tokio I/O reader stuff. It adds 3 dependencies and requires nightly (due to `impl Stream`).

src/lib.rs
```diff
@@ -468,7 +473,43 @@ impl Inotify {
...
+
+        // Use this when 1.24 is available for use. We can use the hard-coded 16 due to Linux's ABI
+        // guarantees.
+        // const EVENT_MAX_SIZE: usize = mem::size_of::<ffi::inotify_event>() + (FILENAME_MAX as usize) + 1;
+        const EVENT_MAX_SIZE: usize = 16 + (FILENAME_MAX as usize) + 1;
+
+        Ok(length_delimited::Builder::new()
+            .little_endian()
+            // FIXME: See https://github.com/tokio-rs/tokio/pull/144
+            // .native_endian()
```
mathstuf: This method doesn't exist yet. See tokio-rs/tokio#144.

src/lib.rs
```diff
@@ -543,6 +558,95 @@ fn read_into_buffer(fd: RawFd, buffer: &mut [u8]) -> isize {
     }
 }
 
+// Use this when 1.24 is available for use. We can use the hard-coded 16 due to Linux's ABI
+// guarantees.
+// const EVENT_MAX_SIZE: usize = mem::size_of::<ffi::inotify_event>() + (FILENAME_MAX as usize) + 1;
+const EVENT_MAX_SIZE: usize = 16 + (FILENAME_MAX as usize) + 1;
```
hannobraun: Hmm, 1.24 is out now, but maybe it's better to leave this as is for the time being. Maybe some people will be stuck with older Rust versions on some Linux distributions. Any opinions? I don't feel strongly either way.

In one of your comments, you mentioned the possibility of making the value of this constant configurable via a Cargo feature, but I think this is fine as it is. It's a sane default, and if people have a problem with it they can file an issue. Better that, than front-loading the complexity for a solution that maybe nobody cares about.

src/lib.rs
```diff
@@ -1036,6 +1140,58 @@ impl<'a> Event<'a> {
...
+    /// [`MOVED_FROM`]: event_mask/constant.MOVED_FROM.html
+    /// [`MOVED_TO`]: event_mask/constant.MOVED_TO.html
+    pub cookie: u32,
+
+    /// The name of the file the event originates from
+    ///
+    /// This field is set only if the subject of the event is a file in a
+    /// watched directory. If the event concerns a file or directory that is
+    /// watched directly, `name` will be `None`.
+    pub name: Option<OsString>,
```
hannobraun: Have you considered avoiding the duplication by making `Event` generic?

Something like this:
``` rust
pub struct Event<S> {
    // ...
    pub name: Option<S>,
}
```

Not sure if the lifetime of the `&'a OsStr` would cause trouble, but I guess it's worth a try.

src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![feature(conservative_impl_trait)]
```
hannobraun: I want to keep this crate on stable, but adding the async stuff behind a Cargo feature would be fine. Once `conservative_impl_trait` lands on stable, we can enable async by default.

src/lib.rs
hannobraun: I don't want to merge this as-is. Do you think it's worth waiting for your PR to be merged? If not we can temporarily duplicate the functionality here.

mathstuf: I didn't expect it to; it does have a `WIP` prefix on the commit message :) .

I'd like to know what the tokio developers think about it at least before reimplementing here.

src/lib.rs
mathstuf: I want it on `stable` as well and that's why I kept the commit doing this bit separate. The second commit is preferred, but the higher barrier is something I'd like to avoid if possible.

src/lib.rs
mathstuf: That does work. Commits squashed and updated to do this.

src/lib.rs
mathstuf: I'd like to leave the minimum version as low as possible to start.

src/lib.rs
hannobraun: Sorry, I realize this is WIP and didn't mean to imply that you wanted to get it merged like this. I just wanted to make my opinion explicit, but should have worded that better.

Okay then, let's wait what happens with your tokio PR before proceeding here.

src/lib.rs
hannobraun: So what do you suggest? Only merge the first three commits and leave the rest for later? Merge everything, but hide it behind a Cargo feature? I think both would be find with me.

src/lib.rs
```diff
@@ -931,10 +1035,10 @@ pub struct Event<'a> {
     /// This field is set only if the subject of the event is a file in a
     /// watched directory. If the event concerns a file or directory that is
     /// watched directly, `name` will be `None`.
-    pub name: Option<&'a OsStr>,
+    pub name: Option<S>,
```
hannobraun: Nice! Glad that worked out.

hannobraun:
Okay then, this is looking really good now. My concerns about the additional overhead have been addressed by keeping the old API as-is. There also seems to be no duplication left, as far as I can tell.

Not quite the unified but still efficient API I originally hoped for, but this is good too, and has the advantage of not being impossible :-)

We still need to decide what to do about `convervative_impl_trait` and wait what happens with your tokio PR, but other than that, this looks good to merge. Thanks again, @mathstuf!

src/lib.rs
mathstuf: How about we use the base 3 commits for now (since it's still pre-1.0, this isn't that much of a big deal). Note that one drawback of the base 3 commits is that it's a busy loop (it always schedules itself again), but the ability to be async is worth it to the projects I'm working on at least.

I'll open a PR with the fourth commit which can be used once `impl Trait` is available. I suppose the better API could be put behind a feature flag. Since the signature is different, do we want the old name to be `event_stream_stable` and reserve `event_stream` for the more idealistic implementation?

hannobraun: Okay, sounds good. Let's just merge the first three.

> Note that one drawback of the base 3 commits is that it's a busy loop

It's probably best to clearly document that on `event_stream`.

> I'll open a PR with the fourth commit which can be used once `impl Trait` is available. I suppose the better API could be put behind a feature flag. Since the signature is different, do we want the old name to be `event_stream_stable` and reserve `event_stream` for the more idealistic implementation?

I don't don't quite follow. If we go ahead as you suggested, we won't need any features at all, because we'll never merge anything that's not stable. Or am I misunderstanding something?

mathstuf: The API is different. I'd expect the base one to be deprecated once the preferable one is available.

mathstuf: This PR now contains just the first three commits and I've updated the docstring as well. The name is still `event_stream`.

hannobraun: Ah, I realize now that we were talking about different things. You were talking about the name of the method, while I thought we were still talking about Cargo features.

The current name, `event_stream`, is fine. When we merge your follow-up PR, we'll just change that method. This is still a `0.x` crate, after all.

PR #103, opened 2018-06-06
feat: Add `Inotify::into_event_stream`

hawkw:
This branch allows an `EventStream` to be created from an `Inotify`
but prevents the `Inotify` from closing the file descriptor when
it's dropped. The created `EventStream` will then be responsible for
closing the file descriptor instead.

This allows the `EventStream` to still be used when the `Inotify` that
created it is out of scope, without leaking the file descriptor.

The behaviour of the existing `Inotify::event_stream` function (which
takes `&mut self` rather than `self`, and leaves the `Inotify`
responsible for closing the fd) is unchanged, so this should be
a non-breaking change.

Signed-off-by: Eliza Weisman <eliza@buoyant.io>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -480,8 +480,22 @@ impl Inotify {
     ///
     /// An internal buffer which can hold the maximum possible size is used.
     pub fn event_stream(&mut self) -> EventStream {
-        EventStream::new(self.fd.clone())
+        EventStream::new(self.fd.clone(), false)
     }
+
+    /// Transform `self` into a `Stream` of events.
+    ///
+    /// This behaves the same as [`event_stream`], but unlike that function,
+    /// the file descriptor will be closed when the _stream_ is dropped, rather
+    /// than when the `Inotify` instance that produced it is dropped. This
+    /// should be used when the `EventStream` should live longer than the
+    /// `Inotify`, such as when it is returned from a function.
+    ///
+    /// [`event_stream`]: struct.Inotify.html#method.event_stream
+    pub fn into_event_stream(self) -> EventStream {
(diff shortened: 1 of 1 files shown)


hawkw:
Some background behind this change: I was using `Inotify::event_stream` in a function that returns a stream. The `Inotify` instance only lives inside of that function, and thus is dropped when it returns. However, dropping the `Inotify` immediately closes the file descriptor, so when I tried to use the returned stream, it immediately fails with `EINVAL`, as the fd has been closed. 

I could have dealt with this issue downstream by calling `Inotify::into_raw_fd` after creating the stream, but then I would have to close the fd manually when the stream is over. I thought making the change here rather than handling it downstream made sense.

hannobraun:
@hawkw Thanks for the contribution! Looks very reasonable.

I think it's worth thinking about if there are better ways to implement `event_stream`, as failing with `EINVAL` isn't very user-friendly. Can't come up with anything right now though. In any case, that's not the point of this PR.

Thanks again! I've made a note to release a new version on crates.io, but it might take until next week before I can get to that.

hannobraun:
@hawkw 
> Yeah, I agree! I have some thoughts on potential improvements to the file descriptor management in this library, and I may open another PR if I can get it working. I wanted to get something in that would at least alleviate the problem in my use-case first, though.

Thoughts and PRs are always welcome! Unfortunately I don't have that many brain cycles left these days to devote to this project, so I mostly rely on contributors for the time being.

hawkw:
@hannobraun I've opened #104 to make the additional changes I had in mind. Sorry for the churn! :)

PR #104, opened 2018-06-07
refactor: Add FdGuard responsible for close-on-drop

hawkw:
This commit improves the close-on-drop behaviour for file descriptors,
by adding a `FdGuard` struct that wraps the file descriptor and is
responsible for closing it when it is dropped. Rather than having an
`Inotify` own an `Arc<RawFd>`, and be responsible for clsoing it,
`Inotify` now owns an `Arc<FdGuard>`, and the `FdGuard` is responsible
for closing the fd.

This allows us to share the file descriptor between an `Inotify` and
an `EventStream`, and close the file descriptor only when the `Arc` is
dropped (i.e., when both the `Inotify` _and_ the `EventStream` are
dropped). This makes the `Inotify::into_event_stream` I added in #103
unnecessary, so I've gone ahead and deprecated it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -86,6 +86,7 @@ use std::hash::{
 };
 use std::io;
 use std::io::ErrorKind;
+use std::ops::Deref;
 use std::os::unix::ffi::OsStrExt;
 use std::os::unix::io::{
     AsRawFd,
@@ -97,6 +98,10 @@ use std::path::Path;
 use std::sync::{
     Arc,
     Weak,
+    atomic::{
+        AtomicBool,
+        Ordering,
+    },
 };
 use std::slice;
 use std::ffi::{
@@ -130,8 +135,7 @@ use libc::{
 ///
 /// [top-level documentation]: index.html
 pub struct Inotify {
-    fd           : Arc<RawFd>,
-    close_on_drop: bool,
+    fd: Arc<FdGuard>,
 }
 
 impl Inotify {
@@ -189,8 +193,10 @@ impl Inotify {
             -1 => Err(io::Error::last_os_error()),
             _  =>
                 Ok(Inotify {
-                    fd           : Arc::new(fd),
-                    close_on_drop: true,
+                    fd: Arc::new(FdGuard {
+                        fd,
+                        close_on_drop: AtomicBool::new(false),
+                    }),
                 }),
         }
     }
@@ -268,7 +274,7 @@ impl Inotify {
 
         let wd = unsafe {
             ffi::inotify_add_watch(
-                *self.fd,
+                **self.fd,
                 path.as_ptr() as *const _,
                 mask.bits(),
             )
@@ -339,7 +345,7 @@ impl Inotify {
             ));
         }
 
-        let result = unsafe { ffi::inotify_rm_watch(*self.fd, wd.id) };
+        let result = unsafe { ffi::inotify_rm_watch(**self.fd, wd.id) };
         match result {
             0  => Ok(()),
             -1 => Err(io::Error::last_os_error()),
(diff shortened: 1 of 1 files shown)


hawkw:
I'm not sure why GitCop has given this branch an ❌, AFAICT all the lines are below the limit?

src/lib.rs
```diff
@@ -480,20 +486,23 @@ impl Inotify {
...
+    /// descriptor management issues that necessitated this function have
+    /// been fixed, it should no longer be necessary, and
+    /// [`Inotify::event_stream`] should be used instead.
     ///
-    /// [`event_stream`]: struct.Inotify.html#method.event_stream
+    /// [`Inotify::event_stream`]: struct.Inotify.html#method.event_stream
+    #[deprecated(
+        since = "0.5.2",
+        note = "use [`Inotify::event_stream`](struct.Inotify.html#method.event_stream) instead."
+    )]
```
hannobraun: I think this deprecation is unnecessary, as this method hasn't been released yet (it only exists in the repository). Would you mind removing it instead?

src/lib.rs
```diff
@@ -522,46 +531,39 @@ impl Inotify {
     ///
     /// [`Inotify`]: struct.Inotify.html
     /// [`close`]: ../libc/fn.close.html
-    pub fn close(mut self) -> io::Result<()> {
+    pub fn close(self) -> io::Result<()> {
         // `self` will be dropped when this method returns. The `Drop`
         // implementation will attempt to close the file descriptor again,
         // unless this flag here is cleared.
-        self.close_on_drop = false;
+        self.fd.close_on_drop.store(false, Ordering::SeqCst);
```
hannobraun: Since `Drop` is no longer implemented for `Inotify`, could you update the comment to note that it refers to the `Drop` implementation of `FdGuard`?

src/lib.rs
```diff
@@ -576,6 +578,59 @@ fn read_into_buffer(fd: RawFd, buffer: &mut [u8]) -> isize {
...
+
+    #[inline]
+    fn deref(&self) -> &Self::Target {
+        &self.fd
+    }
+}
+
+impl Drop for FdGuard {
+    fn drop(&mut self) {
+        if self.close_on_drop.load(Ordering::Relaxed) {
```
hannobraun: Can you please explain the reasoning behind `Ordering::Relaxed`? Won't loading the `bool` that way completely negate the guarantees that storing it with `Ordering::SeqCst` would otherwise have?

src/lib.rs
hawkw: Will do. I wasn't sure what your preferred rules around depredations are, so I figured I'd make the more conservative choice here (also, wanted to avoid breaking my own build :P).

src/lib.rs
hawkw: Ooh, I'm really glad you noticed that. I think this is just a typo on my part. This was supposed to be `Ordering::Release` but my editor must have autocompleted the wrong variant.

Thanks so much for catching that, I totally missed it!

hannobraun:
Thanks for the changes! Everything looks good on first glance. I'll take a closer look as soon as I get the chance.

Those Travis failures look very weird. Not sure if they're related to this PR.

hawkw:
@hannobraun Hmm, it's certainly possible I've broken something. I'll look into them further.

hawkw:
Ah, sorry, that was entirely my fault, fixed it in b06157a115bc6514d13f8d4d196efb667a196f42. I thought the change in 5dc7ebe1f7a732e3deb79cadecd277b901f1b387 was trivial enough that I wouldn't have to SSH into my linux box to rerun the tests, but clearly I was wrong about this. Sorry about that.

hannobraun:
Don't worry, that's why we have Travis! I'm really surprised that you can get this "illegal instruction" error by passing the wrong enum variant. I wonder if that's a bug in Rust.

hawkw:
@hannobraun I believe what happened there is that passing in the wrong variant normally panics with an actually descriptive message saying something to the effect of "releasing on a load is impossible". But, since the load was inside of a `Drop` impl here, it was called while panicking (presumably due to a test failure?), causing a double panic, and double panics kill the program with a SIGILL.

Personally, I think using panics to signal test failure may have been a poor design choice. They cause double panics like that to occur in code that wouldn't normally panic, resulting in a hard to debug error message. But, that's what we've got, so 🤷‍♀️

hannobraun:
@hawkw Ah, makes sense, thanks for the explanation!

I think this looks pretty good, but I still haven't taken a close look at everything. I'll do that as soon as I can find the time.

hawkw:
@hannobraun that's fine, please take your time! I appreciate that you want to give as thorough a review as possible (especially since I introduced a couple bugs in this one 😝)!

hannobraun:
Sorry for the delay. Looks good to me. Thank you, @hawkw!

hawkw:
Excellent! Thanks for the thorough review, @hannobraun!

PR #105, opened 2018-06-07
feat: Implement `EventStream` using `mio`

hawkw:
This branch changes the implementation of `Stream` for `EventStream`
to use `mio` and `tokio_reactor` to register interest in events on
the Inotify file descriptor with the reactor, rather than using
`task::current().notify()` to unpark the task in the executor when
polling the fd would block.

Since the reactor is intended to drive all IO resources (such as the
inotify FD) and to be the bridge between tasks in the executor and
events from the OS, using `PollEvented` is a more correct way to
implement the stream.

This branch shouldn't result in any noticeable changes in behaviour,
but will hopefully allow tokio to manage Inotify's IO events more
intelligently.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,10 +23,13 @@ maintenance = { status = "actively-developed" }
 travis-ci   = { repository = "inotify-rs/inotify" }
 
 [dependencies]
-bitflags    = "1"
-futures     = "0.1"
-inotify-sys = "0.1.1"
-libc        = "0.2"
+bitflags      = "1"
+futures       = "0.1"
+inotify-sys   = "0.1.1"
+libc          = "0.2"
+mio           = "0.6"
+tokio-io      = "0.1"
+tokio-reactor = "0.1"
 
 [dev-dependencies]
 tempdir = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,14 +22,18 @@ categories = ["api-bindings", "filesystem"]
 maintenance = { status = "actively-developed" }
 travis-ci   = { repository = "inotify-rs/inotify" }
 
+[features]
+default = ["stream"]
+stream = ["futures", "mio", "tokio-io", "tokio-reactor"]
+
 [dependencies]
 bitflags      = "1"
-futures       = "0.1"
+futures       = { version = "0.1", optional = true }
 inotify-sys   = "0.1.1"
 libc          = "0.2"
-mio           = "0.6"
-tokio-io      = "0.1"
-tokio-reactor = "0.1"
+mio           = { version = "0.6", optional = true }
+tokio-io      = { version = "0.1", optional = true }
+tokio-reactor = { version = "0.1", optional = true }
 
 [dev-dependencies]
 tempdir = "0.3"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,13 +73,15 @@
 
 #[macro_use]
 extern crate bitflags;
+
+#[cfg(feature = "stream")]
 #[macro_use]
 extern crate futures;
 
 extern crate libc;
 extern crate inotify_sys as ffi;
-extern crate mio;
-extern crate tokio_io;
+
+#[cfg(feature = "stream")]
 extern crate tokio_reactor;
 
 use std::mem;
@@ -113,7 +115,6 @@ use std::ffi::{
     CString,
 };
 
-use futures::{Async, Poll, Stream};
 use libc::{
     FILENAME_MAX,
     F_GETFL,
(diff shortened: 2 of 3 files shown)


hannobraun:
Thanks for the PR! I'll take a closer look once #104 has been merged.

hannobraun:
Now that #104 is merged, this PR is no longer blocked.

@hawkw I'm a bit clueless about mio/futures/tokio, so I'm not too confident in my ability to provide a good review. That said, I can't find anything wrong with this PR.

There's one thing that's keeping me from merging this right now: There's an open PR (#101) that is a follow-up to the initial futures work (#99). It was blocked back then (may still be?), so I couldn't merge right away. It's my impression that your PR supersedes #101. Is that correct?

@mathstuf This PR seems to supersede your open PR #101. Do you have any thoughts on this PR? I'm considering merging this PR and closing #101. Do you see any reason not to do that? Please note that this is based on a previous PR, and the first 4 commits are already merged. As of this writing, only fa6bf8a and 2c63299 are relevant.

hawkw:
First, just FYI, I've rebased this branch onto master, so the diff should no longer include diffs from PR #104.

@hannobraun 
> It's my impression that your PR supersedes #101. Is that correct?

Yeah, that's correct. I haven't looked _too_ closely at #101, but here are some of the important differences from my perspective:
+ It looks like #101 uses the old version of tokio (i.e. the `tokio-core` and `tokio-io` crates), while this branch uses the new `tokio` crate. These are incompatible, and IMO since a majority of the ecosystem has migrated to the new `tokio`, it's better to use that.
+ On master, `Inotify::event_stream` method returns a concrete `EventStream` type. On #101, this changes to `impl Stream<...>`, while on this branch, we still return a named `EventStream` type. Personally, I think that while `impl Trait` is nice to use in applications or in internal code that's not part of a library's public API, it's not ideal for library methods to return an `impl Trait`, as it limits the way users of the library can use that returned type. For example, if I wanted to put the stream returned by `Inotify::event_stream` in a struct field, or use it as a type parameter to some other type, I'd have to box it. This could be a potentially breaking change.
+ #101 depends on the `tokio_file_unix`crate, while my branch doesn't. I'm not familiar with that crate, and this isn't a value judgement, just a difference I noticed.

hannobraun:
Thank you for the explanation, @hawkw, this is great! I'll give @mathstuf a few days to respond before making a decision, but so far everything looks great.

hawkw:
For what it's worth: we just found that switching to this branch from master fixed an issue where `inotify-rs` was causing our application to pin CPU use at 100%.

olix0r:
In particular, [this line](https://github.com/inotify-rs/inotify/pull/105/files#diff-b4aea3e418ccdb71239b96952d9cddb6L685) is what causes this stream to burn CPU, since the executor can't leave this stream as parked (since the task is always notified).

hannobraun:
@olix0r Thank you for the info. If I remember correctly, we decided to land a preliminary version of futures support, as the more efficient version was blocked for various reasons.

Looking back, this seems like a mistake on my part. We should have documented this extensively and hidden it behind a Cargo feature, at the very least. I'm sorry for the trouble this caused you!

olix0r:
@hannobraun No worries! Luckily we caught this in testing ;)

Thanks for putting together this very useful library.

hawkw:
@hannobraun I think releasing the futures support experimentally was a _good_ choice on your part, given that otherwise, we wouldn't have found this issue at all. 

Honestly, I'm mostly just surprised nobody else hit this issue earlier, as I'm fairly sure could occur any time the `EventStream` was used (rather than being triggered by some particular uncommon condition). I guess nobody was using the futures support seriously enough to care about its performance until us.

By the way, thanks for being so willing to work with me on landing all the patches I've submitted, I really appreciate it!

hannobraun:
@olix0r @hawkw Thank you both for your kind words!

> I guess nobody was using the futures support seriously enough to care about its performance until us.

Yeah, probably. Judging from the downloads numbers on crates.io, this library must see a fair bit of use. Unfortunately, not a lot gets back to me.

> By the way, thanks for being so willing to work with me on landing all the patches I've submitted, I really appreciate it!

No problem. Thank *you* for the patches!

hannobraun:
I'm merging this. Sorry for the additional delay.

@hawkw Thank you again for your contributions!

hawkw:
@hannobraun 
>  Sorry for the additional delay.

No problem at all! 

> @hawkw Thank you again for your contributions!

My pleasure! It was fun. :D Thanks for all your reviews.

mathstuf:
> @mathstuf This PR seems to supersede your open PR #101. Do you have any thoughts on this PR?

Sorry, was on an extended vacation. Superseding is fine. This does look more idiomatic than my PR. Thanks @hawkw!

PR #117, opened 2018-08-24
Don't return spurious filenames

dimbleby:
Be careful not to read a name from beyond the valid event length.

Eg I was hitting examples where event.len is zero, but the first byte in the buffer after the event happened to be non-zero.  Then the old code would interpret that as the name.

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -378,3 +378,51 @@ bitflags! {
         const UNMOUNT = ffi::IN_UNMOUNT;
     }
 }
+
+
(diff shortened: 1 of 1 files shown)


hannobraun:
Thank you for the pull request, @dimbleby! I don't have time to take a close look right now, but I'll do so later this week.

hannobraun:
I pushed a regression catch to prevent that bug from showing up again. Thank you again, @dimbleby, good catch!

PR #119, opened 2018-09-05
examples: add an example using tokio's event-loop

ceyusa:
This example shows how to use inotify's stream as future polled by tokio's event loop which is cheap in computational terms using Rust idioms.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ travis-ci   = { repository = "inotify-rs/inotify" }
 
 [features]
 default = ["stream"]
-stream = ["futures", "mio", "tokio-io", "tokio-reactor"]
+stream = ["futures", "mio", "tokio", "tokio-io", "tokio-reactor"]
 
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -35,6 +35,7 @@ futures       = { version = "0.1", optional = true }
 inotify-sys   = "0.1.3"
 libc          = "0.2"
 mio           = { version = "0.6", optional = true }
+tokio         = { version = "0.1", optional = true }
```
hannobraun: Is it possible to move that dependency to `[dev-dependencies]`, or will that cause problems, due to the dependency being optional?

examples/tokio-streams.rs
```diff
@@ -0,0 +1,36 @@
...
+extern crate tempdir;
+extern crate tokio;
+
+use inotify::{Inotify, WatchMask};
+use std::{fs::File, io, thread, time::Duration};
+use tempdir::TempDir;
+use tokio::prelude::*;
+
+/// XXX: hic sunt dracones
+static mut BUFFER: [u8; 32] = [0; 32];
```
hannobraun: You can move the static into the function. Still bad, as it still requires unsafe, but at least it won't be accessible from anywhere else.

hannobraun: Could you also add a comment explaining that this is a workaround for #120, and should be avoided, if possible?

Cargo.toml
ceyusa: yes, it causes problems because of that

hannobraun:
Thanks again, @ceyusa!

PR #122, opened 2018-10-07
Cache downloaded cargo packages

thedrow:
This should speed up the build next time we trigger it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ rust:
   - stable
   - beta
   - nightly
+  
+cache: cargo
 
 script:
   - cargo build --all


thedrow:
If you are experiencing problems with caching, than something is off with Travis.
Downloading from S3 should be very fast.

It was an easy change, I don't mind if we don't merge it.

hannobraun:
> If you are experiencing problems with caching, than something is off with Travis.
Downloading from S3 should be very fast.

I don't think download speed was the problem, but I remember packing and unpacking of caches to take a long time.

thedrow:
Let's close this than.

PR #123, opened 2018-10-08
readme: fix example

srvaroa:
Use EventMask and WatchMask instead of event_mask and watch_mask, which
don't compile.


hannobraun:
Thank you, @srvaroa!

PR #124, opened 2019-01-25
fix: Allow more flexible EventStream buffers

### LINKED ISSUE #120: Stream-based API is difficult to use due to buffer lifetime
See https://github.com/inotify-rs/inotify/issues/106#issuecomment-418343576.

#119 adds an example for working around this issue.

hawkw:
This commit changes the `EventStream` type to hopefully avoid some of
the previously observed difficulties (viz. #120) around the lifetime of
the user-provided buffer.

Previously, calls to `EventStream::new` were required to provide an
`&'a mut [u8]`, where the `EventStream` is generic over the lifetime
`'a`. This is an issue in many cases, where the `EventStream` must
live for the `'static` lifetime, as it requires a `'static mut` buffer,
which is difficult to create safely.

This branch changes the `EventStream` type to be generic over a type
`T` which implements `AsRef<[u8]>` and `AsMut<[u8]>`. Since an
`&mut [u8]` implements these traits, means that the `EventStream` may
now be constructed with an owned buffer _or_ a borrowed buffer. Thus,
users can pass in a `&mut [u8]` to a short-lived `EventStream`, while a
`'static` one can be constructed by passing in a `[u8; N]`, a
`Vec<u8>`, _or_ a `&'static mut [u8]`.

Additionally, this means that buffer types from other crates (such as
`bytes`) could theoretically be used, without requiring `inotify` to
depend on those crates. I'm not sure if this would actually happen that
often in practice, but it's a nice side benefit.

Fixes #120.

Signed-off-by: Eliza Weisman <eliza@buoyant.io>

### DIFF
--- a/examples/issue-120-workaround.rs
+++ b/examples/issue-120-workaround.rs
@@ -9,10 +9,6 @@ use tempdir::TempDir;
 use tokio::prelude::*;
 
 fn main() -> Result<(), io::Error> {
-    /// XXX: this is a workaround to set the inotify's buffer to be
-    /// used, matching with tokio's runtime API
-    /// https://github.com/inotify-rs/inotify/issues/120
-    static mut BUFFER: [u8; 32] = [0; 32];
 
     let mut inotify = Inotify::init()?;
 
@@ -26,7 +22,7 @@ fn main() -> Result<(), io::Error> {
     });
 
     let future = inotify
-        .event_stream(unsafe { &mut BUFFER })
+        .event_stream([0; 32])
         .map_err(|e| println!("inotify error: {:?}", e))
         .for_each(move |event| {
             println!("event: {:?}", event);
--- a/tests/main.rs
+++ b/tests/main.rs
@@ -61,7 +61,7 @@ fn it_should_watch_a_file_async() {
 
     use futures::Stream;
     let events = inotify
-        .event_stream(&mut buffer)
+        .event_stream(&mut buffer[..])
         .take(1)
         .wait()
         .collect::<Vec<_>>();
--- a/src/inotify.rs
+++ b/src/inotify.rs
@@ -402,8 +402,10 @@ impl Inotify {
     ///
     /// [`Inotify::event_stream_with_handle`]: struct.Inotify.html#method.event_stream_with_handle
     #[cfg(feature = "stream")]
-    pub fn event_stream<'buffer>(&mut self, buffer: &'buffer mut [u8])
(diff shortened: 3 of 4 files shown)


examples/issue-120-workaround.rs
```diff
@@ -26,7 +22,7 @@ fn main() -> Result<(), io::Error> {
     });
 
     let future = inotify
-        .event_stream(unsafe { &mut BUFFER })
```
hannobraun: I think this example should be removed and this technique here should be part of the API documentation (as a small example). Not necessary to do it in this PR, but we shouldn't forget to open an issue, at least.

tests/main.rs
```diff
@@ -61,7 +61,7 @@ fn it_should_watch_a_file_async() {
 
     use futures::Stream;
     let events = inotify
-        .event_stream(&mut buffer)
+        .event_stream(&mut buffer[..])
```
hannobraun: I assume this change was required to get it to compile? If so, this is a breaking change. Not a problem, just want to confirm.

hannobraun:
@hawkw 
Since you seem to be using this in production, would you like me to do a new (minor) release right now?

I'd prefer to let this sit for a while, due to laziness/busyness, to let it mature for a bit, and potentially to group it with other breaking changes that might be coming. But I'm happy to cut a release right now, if that helps you out.

tests/main.rs
hawkw: Yeah, that's correct --- it's only necessary when the array is longer than `[T; 32]`, as only arrays of length 0-32 have `AsRef`/`AsMut` impls.

hawkw:
@hannobraun if you're anticipating more breaking changes in the near future, I think we're fine with a Git dependency temporarily. 

Alternatively, I think it may be possible to make the change more backwards-compatible, by adding a new constructor that takes a `T: AsRef + AsMut`, leaving `new`'s signature as it was previously, and having it slice the buffer and pass it to the new constructor. However, I think this is probably not a great idea, as it results in a more confusing API, and the `EventStream` type's generic parameter is still changing, so this is still a breaking change when the `EventStream` is e.g. a struct field...

hannobraun:
@hawkw I'm not anticipating more breaking changes, I just figured we might decrease some churn in case any more are coming. Sounds like a new release is the way to go though. I'll try to get it done soon.

I agree that attempts to alleviate the breakage are probably counter-productive, for the reasons you mentioned.

PR #131, opened 2019-06-22
Format code using 'cargo fmt'

Atul9:


hannobraun:
Thank you for this pull request, @Atul9, but I'm not going to merge it.

`cargo fmt` provides consistent formatting. I don't care about consistent formatting, I care about readable formatting, and I don't think these changes make the code more readable. I don't think they make it particularly less readable either, but in the absence of any clear advantage, I'd rather not introduce a lot of churn that will introduce merge conflicts with any ongoing work.

I realize that a lot of people value consistent formatting highly, and they seem to believe that the reasons why consistent formatting is good are self-evident. Well, they are not evident to me. As I said, I value readable code, and while consistency is a big part of achieving that, I've found that formatting all cases the same can often lead to less readable code.

I'd be happy to change my opinion, so if you have any good arguments on why running `cargo fmt` on the code is an improvement, please share them!

Thanks again!

Issue #121, opened 2018-09-25
Async/Await syntax support


thedrow:
Futures 0.3 will support the Async/Await syntax.
One it reaches GA, Tokio will be next.
We should investigate what it will require of us to support the new syntax and prepare ourselves for it.

thedrow:
Seems like 0.3 alpha 3 has a compatibility layer with futures 0.1 that includes Tokio compatibility.
Maybe we should try that first?

hannobraun:
Thank you for opening this issue! 

I don't have the time to look into it right now (and likely won't have time for a while), but I agree that this is important. I'm happy to assist by reviewing reviewing pull requests or with general feedback.

Side note: There's [a problem](https://github.com/inotify-rs/inotify/issues/120) with the current stream-based API. Something to keep in mind when updating or redesigning the API.

hannobraun:
We've had support for async/await for a while now. Closing.

PR #134, opened 2019-09-18
feat: update to tokio 0.2 and futures 0.3

sopium:
(Stability note: this feature depends on nightly rust and alpha releases of tokio 0.2 and futures 0.3. Maybe you want to hold off the merge until they are stable. In which case I will try to make sure that this PR is up to date.)

This is a minimal and straightforward update to make inotify up to date with `std::task`, async-await syntax and next versions of tokio and futures.

## API changes

* `EventStream` is now a futures 0.3 `Stream`

  A futures 0.3 `Stream` can be used with async-await directly (cc #121). See the stream example.

* The buffer passed to the `event_stream` method is now required to be `Unpin`

## Other changes

+ Mandatory dependencies is minimized: only `tokio-net`, `tokio-io` and `futures-core` are mandatory dependencies. And for `tokio-net` unused features are disabled. The full `tokio` crate and `futures-util` are dev-dependencies.

+ The crate is updated to 2018 edition, so that related tests and
  examples can be written in async/await.

+ The CI config is modified to:
  * Run on nightly
  * Run with `--no-default-features` as well


Cargo.toml
```diff
@@ -26,21 +27,22 @@ travis-ci   = { repository = "inotify-rs/inotify" }
...
-tokio         = { version = "0.1", optional = true }
-tokio-io      = { version = "0.1", optional = true }
-tokio-reactor = { version = "0.1", optional = true }
+bitflags             = "1"
+futures-core-preview = { version = "=0.3.0-alpha.18", optional = true }
+inotify-sys          = "0.1.3"
+libc                 = "0.2"
+mio                  = { version = "0.6", optional = true }
+tokio-io             = { version = "=0.2.0-alpha.4", optional = true }
+tokio-net            = { version = "=0.2.0-alpha.4", optional = true, default-features = false }
```
hannobraun: Some of the versions are pinned to an exact alpha version. Is there a specific reason for that?

I think requiring an exact version should usually be avoided, as that can cause problems with other crates that might have slightly different version requirements.

Cargo.toml
```diff
@@ -26,21 +27,22 @@ travis-ci   = { repository = "inotify-rs/inotify" }
...
 
 [dev-dependencies]
-tempdir = "0.3"
+tempdir              = "0.3"
+futures-util-preview = "=0.3.0-alpha.18"
+tokio                = "=0.2.0-alpha.4"
```
hannobraun: Same as above. Versions are pinned.

src/inotify.rs
```diff
@@ -405,7 +405,7 @@ impl Inotify {
     pub fn event_stream<T>(&mut self, buffer: T)
         -> EventStream<T>
     where
-        T: AsMut<[u8]> + AsRef<[u8]>,
+        T: AsMut<[u8]> + AsRef<[u8]> + Unpin,
```
hannobraun: I'm a bit concerned about the `Unpin`, as it makes the API less flexible. I think right now one can pass an array, but with the `Unpin`, you'd either need a slice (possibly causing lifetime issues) or a heap allocation.

I'm not sure if it can be avoided and if that's a big deal in practice. I just figured I'd point out my concerns, in case you have something informative to say :-)

src/inotify.rs
```diff
@@ -422,7 +422,7 @@ impl Inotify {
     pub fn event_stream_with_handle<T>(&mut self, handle: &Handle, buffer: T)
         -> io::Result<EventStream<T>>
     where
-        T: AsMut<[u8]> + AsRef<[u8]>,
+        T: AsMut<[u8]> + AsRef<[u8]> + Unpin,
```
hannobraun: See comment above about `Unpin`.

src/stream.rs
```diff
@@ -47,7 +39,7 @@ pub struct EventStream<T> {
 
 impl<T> EventStream<T>
 where
-    T: AsMut<[u8]> + AsRef<[u8]>,
+    T: AsMut<[u8]> + AsRef<[u8]> + Unpin,
```
hannobraun: See comment above about `Unpin`.

src/stream.rs
```diff
@@ -78,36 +70,37 @@ where
 
 impl<T> Stream for EventStream<T>
 where
-    T: AsMut<[u8]> + AsRef<[u8]>,
+    T: AsMut<[u8]> + AsRef<[u8]> + Unpin,
```
hannobraun: See comment above about `Unpin`.

src/stream.rs
```diff
@@ -78,36 +70,37 @@ where
...
 {
-    type Item = EventOwned;
-    type Error = io::Error;
+    type Item = io::Result<EventOwned>;
 
-    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error>
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>>
     {
-        if self.unused_bytes == 0 {
+        let self_ = self.get_mut();
```
hannobraun: Can we use [`get_unchecked_mut`](https://doc.rust-lang.org/std/pin/struct.Pin.html#method.get_unchecked_mut) here, to forgo the requirement for `Unpin`? I don't think we're moving out of `self_`, so it should be fine, I think.

hannobraun:
Two more things:
- Note to my future self: Travis didn't run for some reason. I'll have to figure out why, if it doesn't fix itself at some point.
- @sopium:  For future reference, I'd prefer to have each self-contained change  (e.g. the update to Rust 2018) in its own commit. You don't need to change this pull request (it's not a big change anyway), but I'd appreciate it if future pull requests had smaller commits, as those are easier to review.

Cargo.toml
sopium: These crates might have breaking changes between different alpha versions, so they are pinned to a specific version. It won't be necessary once they are stable.

src/inotify.rs
sopium: Arrays *are* `Unpin`. But your general point is valid.

Cargo.toml
hannobraun: Makes sense.

src/inotify.rs
hannobraun: Interesting, I didn't think they were!

hannobraun:
Thanks for the change, @sopium! Let's hope things become stable soon, so we can merge this. Until then, more reviews/comments are always welcome.

peku33:
Any updates on this? Waiting for merge! Maybe this could become available as pre-release?

hannobraun:
@peku33 My understanding is, that this still requires nightly. I don't want to merge it as long as this is the case, as I want the master branch to always be in a releasable state (if someone submits a bugfix, for example).

Would it be practical for you to use a Git dependency in Cargo, to depend on @sopium's branch directly?

peku33:
@hannobraun How about doing the same as tokio itself did?
At the moment we have `tokio-0.1.22` which is considered stable. We also have `tokio-0.2.0-alpha.6` (as of today) which is a pre-release, IIRC requires rust nightly, futures-preview (also alpha). AFAIK, when you use `cargo add tokio` it will add `0.1.22`. But running `cargo add tokio --allow-prerelease` will install `0.2.0-alpha.6` version. This allows users to choose stable/nightly version pretty easily.
I think keeping current version as stable is reasonable, but why not publish a pre-release (eg `0.8.0-alpha.1` or `1.0.0-alpha.1`)?
Current codebase looks pretty stable, since inotify itself won'y probably evolve in the nearest future, there wouldn't probably be more development against current branch.

hannobraun:
@peku33 
> I think keeping current version as stable is reasonable, but why not publish a pre-release

Because that would take time to do and I don't have much of that. From my perspective, it's a much more economical decision to just wait until everything is stable, then merge and release.

peku33:
Hmm, pretty sad, especially taking into account fact, that lots of popular crates uses such approach (tokio itself, mysql_async etc).

@hannobraun Are you sure accepting @sopium PR with version set to `0.8.0.alpha-1` is too much? This would *not* automatically update all clients to this unstable version. Even the main version on crates.io etc will still be the current *stable* version (take a look at https://crates.io/crates/mysql_async ) ;)

peku33:
@sopium Any way to use event_stream with buffer larger than 32 bytes? For now it looks like `AsMut<[u8]>` accepts arrays of up to 32 bytes...

sopium:
> 
> 
> @sopium Any way to use event_stream with buffer larger than 32 bytes? For now it looks like `AsMut<[u8]>` accepts arrays of up to 32 bytes...

You can use `Vec<u8>`, `&mut buf[..]`, or define a newtype wrapper:

```rust
struct LargeU8Array([u8; 512]);

impl AsRef<[u8]> for LargeU8Array {...}

impl AsMut<[u8]> for LargeU8Array {...}
```

hannobraun:
@peku33 
> Hmm, pretty sad, especially taking into account fact, that lots of popular crates uses such approach (tokio itself, mysql_async etc).

I guess those crates have less busy maintainers then, because yes, I am sure that adding yet another thing to my plate would be too much.

You haven't commented on my suggestion to add a Git dependency on @sopium's branch in Cargo. Doing so would require no additional work from anyone.

If that's not good enough for some reason (the only one I can think of would be that you want to publish a crate with that dependency to crates.io), you can publish your own version of inotify. Call it `peku33-inotify` or `inotify-async-nightly` or whatever. Then anyone else who can't wait for the next version of `inotify` can use that in the meantime.

sopium:
The latest commit is in sync with tokio master.

In tokio master, `PollEvented::new` now returns `io::Result<_>`, and `PollEvented::new_with_handle` is removed. As a result, the `event_stream` API has changed accordingly.

If you are still using `tokio 0.2.0-alpha.6`, stick to the previous commit.

emarcotte:
I'm not exactly fluent in all the ins/outs of the new async stuff. Is it reasonable to implement EventStream as a FusedStream to be more compatible with select!()? E.g. https://rust-lang.github.io/async-book/06_multiple_futures/03_select.html

I'm not sure if I should be wrapping it externally or if the stream itself should track a terminated flag... Still fooling around with it but does something like this make sense?

```
diff --git a/src/stream.rs b/src/stream.rs
index e9e56c3..38f9976 100644
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -12,6 +12,7 @@ use mio::{
 };
 use tokio::io::AsyncRead;
 use futures_core::{Stream, ready};
+use futures_core::stream::FusedStream;
 use tokio::net::util::PollEvented;
 
 use crate::events::{
@@ -32,6 +33,7 @@ pub struct EventStream<T> {
     buffer: T,
     buffer_pos: usize,
     unused_bytes: usize,
+    finished: bool,
 }
 
 impl<T> EventStream<T>
@@ -45,10 +47,20 @@ where
             buffer: buffer,
             buffer_pos: 0,
             unused_bytes: 0,
+            finished: false,
         })
     }
 }
 
+impl<T> FusedStream for EventStream<T>
+where
+    T: AsMut<[u8]> + AsRef<[u8]>,
+{
+    fn is_terminated(&self) -> bool {
+        self.finished
+    }
+}
+
 impl<T> Stream for EventStream<T>
 where
     T: AsMut<[u8]> + AsRef<[u8]>,
@@ -69,6 +81,7 @@ where
         if self_.unused_bytes == 0 {
             // The previous read returned `0` signalling end-of-file. Let's
             // signal end-of-stream to the caller.
+            self_.finished = true;
             return Poll::Ready(None);
         }
 
```

sopium:
Updated to tokio v0.2.1.

The `event_stream_with_handle` method is removed becuase `PollEvented` no longer has `new_from_handle`.

hannobraun:
Thank you @sopium for your continued work on this! The changes look good to me, but for some reason, the CI build doesn't run. I've set up GitHub Actions instead. Can you please rebase on master, so it runs for this pull request?

hannobraun:
Thank you, @sopium! I'll try to release a new version soon.

Issue #112, opened 2018-08-08
Consider unifying traditional and future-based APIs


hannobraun:
Currently there are two parallel APIs: The original one, and the new future-based one. I think it would be ideal if we could unify those.

One problem I see is efficiency. The stream-based API requires one additional heap allocation for every event that has a `name`. The reason for this are lifetime issues that probably can't be resolved, at least not in safe Rust.

I think it's best for now to keep things as they are, while keeping a look at how futures developers. Feedback is very welcome!

calmofthestorm:
One thought on this would be that the current stream API takes &mut on the Inotify and does not provide a way to add additional watches. I'm working on something where I want to recursively watch a directory, and thus need to add new watches on any newly created directories. One solution would be to allow mutably borrowing an inotify from the stream, or adding functions to add/remove watches directly.

Personally I would favor adding a method to clone an Inotify. It should be safe (in a Rust sense) and the underlying representation is already friendly to this. I understand this would permit certain undesirable uses (e.g., two streams on the same inotify would each get a subset of events).

I've tried a few ways of writing this involving multiple threads, etc, which have all been clumsy. Next I want to try using `select` along with `read_events` which I think should be simpler.

hannobraun:
Thank you for your feedback, @calmofthestorm. I'm very open to improving the situation and make your use case possible, but I think allowing users to clone `Inotify` would be taking it too far. It would be very error-prone, for the reasons you mention. Adding more methods to `EventStream` would certainly be acceptable. Maybe there could be a separate API, `Watches`, with `add` and `remove` methods that can be accessed both through `Inotify` and `EventStream` (e.g. `inotify.watches().add(...)`). Maybe it could even be possible to clone a `Watches` and use it anywhere.

But does creating an `EventStream` even prevent you from using the original `Inotify` instance? `Inotify::event_stream` takes `&mut self`, but the resulting `EventStream` doesn't keep that reference. That means it should already be possible to use an `Inotify` and an arbitrary number of `EventStream`s in parallel, which seems wrong to me.

Maybe it would be better to replace `Inotify::event_stream` with an `into_event_stream` method that consumes the `Inotify`, and make sure `EventStream` has all the methods that users need.

calmofthestorm:
Thanks, I missed that the mut reference is not saved by `EventStream` (but I agree with your concern that this is wrong).

[...]

Issue #111, opened 2018-08-07
Multiple `EventStream`s can take each other's events


hannobraun:
It is possible to create multiple `EventStream`s from a single `Inotify` instance. Those streams will `read` from the same inotify file descriptor. Depending on the timing of the reads, one `EventStream` could end up with all the events, or both could end up with some of them. (Note: I haven't tested this. It's just my own reasoning).

It would be possible to guarantee there is only one `EventStream` per `Inotify` instance, by consuming or mutably borrowing `Inotify` when creating an `EventStream`. However, both those solutions would make `add_watch`/`rm_watch` inaccessible.

I have two ideas for how to solve this:
1. Duplicate `add_watch`/`rm_watch` on `EventStream`. That way, `EventStream` would effectively become a replacement for `Inotify` that uses a different method for reading events.
2. Split `Inotify` into two parts, one for managing watches, one for reading events.

I don't think solution 1 is flexible enough. `Stream` has a lot of methods that move the stream, which would make `add_watch`/`rm_watch` inaccessible again. This isn't really a solution, it's just kicking the problem one step down the line.

Solution 2 deserves a bit more elaboration:
- `Inotify` would become a struct with two fields, `watches` and `events`.
- `watches` would provide `add_watch`/`rm_watch` (which could be shortened to `add`/`remove`).
- `events` would provide the various `read_*` methods.
- `events` could be moved or borrowed when creating an `EventStream`, without affecting `watches`.

Right now I'm unsure what to do, although I like solution 2. I'm going to let this issue sit for a while before attempting an implementation. In the meantime, feedback is highly welcome.

talklittle:
This issue is the same as #176 and #177

hannobraun:
Thanks for letting me know, @talklittle! Closing in favor of those issues.

PR #135, opened 2019-09-18
Async await support

### LINKED ISSUE #121: Async/Await syntax support

sopium:
(This includes changes from #134.)

This adds an async `Inotify` type. It has an `async fn read_events` (close #121):
```rust
    pub async fn read_events<'a>(&mut self, buffer: &'a mut [u8])
                           -> io::Result<Events<'a>>
```
that is basically the same as the original `read_events` function, but awaits on the `read`:
```rust
    {
        let num_bytes = self.fd.read(buffer).await?;

        if num_bytes == 0 {
            return Err(
                io::Error::new(
                    io::ErrorKind::UnexpectedEof,
                    "`read` return `0`, signaling end-of-file"
                )
            );
        }

        Ok(Events::new(Arc::downgrade(self.fd.get_ref()), buffer, num_bytes))
    }
```

This addresses the efficiency concern in #112: no additional allocation is required.


sopium:
This type also will not have the problems described in #111.

hannobraun:
@sopium Thank you, this looks awesome. I haven't taken a close look yet, but I will once #134 is closer to being merged (feel free to remind me, if I forget).

Once this is merged, do you think there's any reason to keep the original blocking API around?

cc @hawkw (I already pinged you on #134. I figured maybe you're interested in this one too.)

sopium:
I tried to modify the `Inotify` type to make it support both async and blockink API. I think it can work.

The trick is to use a fake `PollEvented` wrapper when the `async-await` feature is not enabled.

hannobraun:
Interesting. What if we removed the `async-await` feature and included that by default, would there be a reason to keep the blocking API? Not sure how the new futures and async/await work in detail, but I think I remember that the older futures hat a `wait` method that could be used to block on a future. Maybe something like that could be used to completely replace the blocking API.

Not sure if that's the right approach. Just wondering.

sopium:
> Interesting. What if we removed the `async-await` feature and included that by default, would there be a reason to keep the blocking API? Not sure how the new futures and async/await work in detail, but I think I remember that the older futures hat a `wait` method that could be used to block on a future. Maybe something like that could be used to completely replace the blocking API.
> 
> Not sure if that's the right approach. Just wondering.

It should be possible: you can create a `current_thread` runtime and `block_on` the future.

But it's quite inefficient, a new runtime need to be created for every blocking read call.

And it means that tokio is always needed. Quite heavy a dependency when you only want to use blocking IO.

hannobraun:
Sounds like it make sense to keep the traditional blocking API then. Thanks for the explanation!

hannobraun:
This pull request is very old, and async/await support has been available in inotify-rs for a long time now. Closing.

PR #157, opened 2020-11-04
Inotify::read_events handles only a buffer's worth of events.

jimblandy:
Change the documentation of `Inotify::read_events` to match the actual behavior:
return an iterator over only as many events can fit in the supplied buffer, not
over any available events. Since this is a backwards-incompatible change, bump
the crate version number to 0.9.0.

It would be possible to make `Events` behave as documented, but it would need to
hold a strong reference to `fd` and a mutable reference to `buffer`, and it
seems awkward. The `notify` crate's use of `read_events` is easily switched by
requesting level-sensitive notification on the inotify file descriptor, which is
designed for exactly this sort of situation.

### DIFF
--- a/src/inotify.rs
+++ b/src/inotify.rs
@@ -295,12 +295,12 @@ impl Inotify {
         result
     }
 
-    /// Returns any available events
+    /// Returns one buffer's worth of available events
     ///
-    /// Returns an iterator over all events that are currently available. If no
-    /// events are available, an iterator is still returned. If you need a
-    /// method that will block until at least one event is available, please
-    /// consider [`read_events_blocking`].
+    /// Reads as many events as possible into `buffer`, and returns an iterator
+    /// over them. If no events are available, an iterator is still returned. If
+    /// you need a method that will block until at least one event is available,
+    /// please consider [`read_events_blocking`].
     ///
     /// Please note that inotify will merge identical unread events into a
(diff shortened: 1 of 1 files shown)


hannobraun:
Thank you, @jimblandy, good catch there! I wonder if the behavior changed at one point, or if the documentation was always wrong. Oh well.

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 
 name    = "inotify"
-version = "0.8.3"
+version = "0.9.0"
```
hannobraun, line 4: Can you please remove this change? I prefer to update the version right before the release, for the following reasons:
- Changing this in pull requests is not a reliable process, as not all contributors do that. We could make it a requirement, but that would put more of a burden on contributors, and also on me, because I'd have to remember to keep track of that.
- Changing the version makes the process of using `[patch]` in Cargo.toml to replace the latest `inotify` release with a Git version more complicated, as `[patch]` only works if the versions are the same.

hannobraun (changes requested):

Cargo.toml
jimblandy: Sure thing.

hannobraun: Thank you!

hannobraun (approved):

PR #161, opened 2020-11-08
Fix task wake up

### LINKED ISSUE #160: EventStream not working
Hello. My apologies - while upgrading to tokio 0.3, I introduced a bug which causes some problems with reading events from an EventStream instance. Only one or two events get read, the rest becomes lost. 

Minimal reproducible example, try this with inotify 0.8.3 and inotify 0.9.0. Move your mouse or type on your keyboard and watch the output of this program. 
```
#[tokio::main]
async fn main() {
    let mut inotify = Inotify::init().unwrap();
    inotify
        .add_watch("/dev/input", WatchMask::ALL_EVENTS)
        .unwrap();

    let mut stream = inotify.event_stream([0; 512]).unwrap();
    while let Some(event) = stream.next().await {
        let event = event.unwrap();
        println!("{:?}", event);
    }
}
```
I suggest that you yank the 0.9.0 release from crates.io until a fix is made. I think it will be something trivial, probably a missing task wake up or something like that.

htrefil:
Wakes up the reading task when a WouldBlock error is encountered. Fixes #160.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -98,7 +98,10 @@ fn read(fd: &AsyncFd<ArcFdGuard>, buffer: &mut [u8], cx: &mut Context) -> Poll<i
 
     match result {
         Ok(read) => Poll::Ready(Ok(read)),
-        Err(ref err) if err.kind() == io::ErrorKind::WouldBlock => Poll::Pending,
+        Err(ref err) if err.kind() == io::ErrorKind::WouldBlock => {
+            cx.waker().wake_by_ref();
+            Poll::Pending
+        }
         Err(err) => Poll::Ready(Err(err)),
     }
 }


hannobraun:
Thank you, @htrefil! No worries, bugs happen. Thank you for providing such a quick fix.

It would be great to have this covered by the test suite (it ran perfectly fine despite this bug). I will merge this now, but if you're in the mood for writing some tests, a follow-up pull request would be very welcome.

htrefil:
I'm not sure it would make sense to make a test case for this since it's a rather specific condition. Async is tricky and next time it might be something quite different.

hannobraun:
I agree that is probably doesn't make sense to add a test for this specific condition, but it might be possible to broaden the test suite, so it covers a larger range of possible bugs. Such broad tests wouldn't be very useful to determine what exactly broke, but they would increase the confidence that changes don't break anything fundamental.

Issue #151, opened 2020-07-24
Consider reverting #146 at some point


lnicola:
The compiler [dropped support](https://github.com/rust-lang/rust/issues/62516) for RHEL 5, so #146 can in principle be reverted when the MSRV gets bumped to 1.47.

hannobraun:
Thanks for the note!

Do you think there's any pressure to revert that change? To me it seems like it doesn't make any difference in practical terms. But I'm happy to change it back, if it's causing problems for anyone.

lnicola:
Nope, it's just cleaning up a workaround that's no longer needed. The two extra syscalls probably don't matter, but there's no point in having them. I filed this to keep track of it.

hannobraun:
Makes sense, thanks again!

PR #168, opened 2021-03-17
Don't fcntl on an invalid fd.

rbartlensky:
Hello! I've stumbled upon an error while using this crate

```
thread 'main' panicked at 'Failed to initialize inotify.: Os { code: 9, kind: Other, message: "Bad file descriptor" }'
```

and I figured it's probably because the code doesn't check whether `inotify_init` returns `-1`.

Let me know if I need to change anything else!

### DIFF
--- a/src/inotify.rs
+++ b/src/inotify.rs
@@ -100,21 +100,20 @@ impl Inotify {
         // https://github.com/rust-lang/rust/issues/12148
         let fd = unsafe {
             let fd = ffi::inotify_init();
+            if fd == -1 {
+                return Err(io::Error::last_os_error());
+            }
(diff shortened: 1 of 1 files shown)


hannobraun:
Thank you, @rbartlensky, good catch!

I think an even better solution would be to remove the `fcntl`s completely and use `inotify_init1` instead. Using `inotify_init` and `fcntl` was a workaround for a downstream issue that is no longer required (see #146 and #151 for context). However, this PR is a clear improvement, so I'll merge it.

If you (or anyone else reading this) wants to address #151 in a follow-up PR, that would be very welcome.

PR #169, opened 2021-04-13
Clarify that Event may have a name for indirectly-watched directories.

calmofthestorm:
Clarify that events will populate the Name field when the Event is in a watched directory but does not concern a file or directory that is watched directly. The previous documentation implied this only applied to files, but at least on 5.10.0-3 I'm finding that it applies to both files and directories, with the relevant point being only whether it is watched directly, and this behavior makes more sense anyway IMO.

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -98,7 +98,7 @@ pub struct Event<S> {
 
     /// The name of the file the event originates from
     ///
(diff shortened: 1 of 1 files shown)


hannobraun:
Thank you, @calmofthestorm!

Issue #155, opened 2020-10-30
Unsoudness: reference misalignement


a1phyr:
Currently, events are read from a `&mut [u8]` buffer, which has an alignment of `1`, whereas `struct inotify_event` has an alignement of `4`.

Therefore, creating a `&inotify_event` from such a `&[u8]` (as done [here](https://github.com/hannobraun/inotify/blob/b3a7062a732851ac39dfd63ba2bd337dbcff123e/src/events.rs#L156-L162)) may create a misaligned reference, which is undefined behaviour.

hannobraun:
Thank you for pointing this out, @a1phyr!

hannobraun:
Presumably fixed by #171, so I'm going to close. Please note that the fix might not be optimal, and any work on trying to provide a more robust fix would be very appreciated. Please check out [this comment](https://github.com/hannobraun/inotify-rs/pull/171#pullrequestreview-655632788) for more information.

PR #171, opened 2021-05-08
Align input buffers before reading or writing inotify events

FenrirWolf:
Attempts to solve #155

This passes the current tests and eliminates the `miri` warning about unaligned references that currently triggers [here](https://github.com/hannobraun/inotify/blob/b3a7062a732851ac39dfd63ba2bd337dbcff123e/src/events.rs#L156-L162), but I'm not certain if it's fully robust. The alignment code works by splitting away any unaligned portion of the input buffer, and so I'm not sure if the byte-tracking code in `Stream` and other places needs to be updated to reflect that.

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -12,6 +12,7 @@ use inotify_sys as ffi;
 
 use crate::fd_guard::FdGuard;
 use crate::watches::WatchDescriptor;
+use crate::util::align_buffer;
 
 
 /// Iterator over inotify events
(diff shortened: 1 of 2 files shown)


hannobraun (approved):
I've checked the rest of the code, and I think everything's fine. The stream code uses the same core primitives that this PR updates (`Event::from_buffer`, `read_into_buffer`), so I think it's fully covered by this fix.

What I don't like about this fix is how subtle it is. The alignment needs to be considered in multiple places, and while I think this PR gets it right, future changes might mess it up again. Still, I think this PR fixes the unsoundness, so I'm going to merge it.

I think it would be better, if we could somehow use a custom buffer type to enforce the alignment at the type level. I'm not sure how practical that is, and I don't have the bandwidth to look into that. If anyone wants to look into that approach, that would be very welcome.

Thanks again, @FenrirWolf!

PR #173, opened 2021-05-12
Check fcntl result

erickt:


hannobraun (approved):
Thank you, @erickt, looks good!

Issue #178, opened 2021-09-07
Implement ToOwned for `Event<&'a OsStr>`


i509VCB:
I am using inotify with calloop and one issue I've noticed is the lack of ability to easily convert an event from the methods into an owned event to pass to calloop. Calloop gets kinda messy when an `Event` in an event source has a lifetime, so getting an owned copy is typically the best way around this issue.

I noticed the presence of `into_owned`that is only available with the streams api, I think we could convert this to delegate to an implementation of `ToOwned` and made always available.

hannobraun:
Hi @i509VCB, thank you for your pull requests! I'm currently on vacation, but I'll be back next week and will take a look then (or maybe the week after, if there's too much of a backlog to catch up on).

i509VCB:
Ah no worries

hannobraun:
#179 has been merged, which is a partial resolution to this issue.

talklittle:
I was curious about this, and from [this StackOverflow question](https://stackoverflow.com/questions/72105604/implement-toowned-for-user-defined-types) it sounds like it is not really needed to implement the `ToOwned` trait unless there is a specific use case that needs that trait in particular. Having the `Event.to_owned()` non-trait method may be sufficient, if I understand correctly.

I am not familiar with the calloop crate, but a search over its [GitHub source](https://github.com/Smithay/calloop) does not find any references to the `ToOwned` trait.

i509VCB:
So in the case of calloop, I need to have an owned event since an event source in calloop has an Event assoicated type. If the assoicated type has a lifetime, then it gets quite ugly fast.

talklittle:
The `Event.to_owned()` method already satisfies that need to get an owned Event, right? I'm having trouble understanding why the `ToOwned` trait should be implemented in addition to that?

Reading the SO question again, it sounds like `ToOwned` isn't really a fit for `Event`, as its intent is to be applied to reference types like `str` and `OsStr` to get a String and OsString. `Event` on the other hand is not a reference type. We can still define `Event.to_owned()` as a convention as a lightweight clone, but that doesn't imply we must implement `ToOwned`.

Maybe the proper solution here is to implement Clone ourselves (just rename `to_owned()` to `clone()` essentially), and then we can use the built-in ToOwned impl for Clone types. And then delete our `to_owned()` method as it would delegate to `clone()`.

talklittle:
[...]

PR #179, opened 2021-09-08
Make `Event::into_owned` always available

i509VCB:
This allows `Event::into_owned` to be used independent of the `stream` feature being enabled. Only a partial resolution to #178 since a future rust version will be needed to implement `ToOwned` since `Event` is Clone


hannobraun (approved):
Thank you, @i509VCB, this is great!

PR #183, opened 2021-10-02
Add Ord/PartialOrd instances to WatchDescriptor

RaoulHC:
As inotify watch descriptors don't have an order per se, an alternative would be to make the `id` field public outside of the crate, so users of this library can implement a wrapper that implements traits based on `id`. Otherwise some data structures can't use a `WatchDescriptor` as a key which can be useful (i.e. `BTreeMap<WatchDescriptor, PathBuf>`).

### DIFF
--- a/src/watches.rs
+++ b/src/watches.rs
@@ -3,6 +3,7 @@ use std::{
         Hash,
         Hasher,
     },
+    cmp::Ordering,
     os::raw::c_int,
     sync::Weak,
 };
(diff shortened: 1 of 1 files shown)


hannobraun (approved):
Thank you, @RaoulHC, looks good to me!

I'd be happy to accept a PR that exposes the id as an additional measure, if required, but I think that this change is fine as-is.

PR #187, opened 2021-10-29
Increase the buffer size for the stream example

### LINKED ISSUE #186: Consider increasing the buffer size for the stream example
Currently, the [buffer size of the stream example](https://github.com/hannobraun/inotify-rs/blob/master/examples/stream.rs#L31) is set to `let mut buffer = [0; 32];`. When I tried to run the example, I hit a rather cryptic OS error, `Invalid Argument`. This was easily fixed by increasing the buffer size.

The [unit test uses](https://github.com/hannobraun/inotify-rs/blob/master/tests/main.rs#L226) `let mut buffer = [0; 1024];`. Maybe the example should use `1024` as well?

It is unlikely that `tempfile` would produce a path in either case that would cause the inotify message to exceed `1024`. I've come up with a function though that given a filepath it will calculate the maximum inotify message size one could expect.

Please, let me know if either of these would be worth opening an MR for.

```rust
fn get_buffer_size(path: &Path) -> usize {
    // Get the length of the parent path if the filepath is not the root directory.
    // Because we canonicalized the filepath above, we do not need to worry about
    // prefixes.
    let name_max = if let Some(parent_path) = path.parent() {
        parent_path.as_os_str().len()
    } else {
        0
    }
    // Add one to account for a `/`, either in between the parent path and a filename
    // or for the root directory.
    + 1
    // Add the maximum number of chars in a filename, 255
    // https://github.com/torvalds/linux/blob/master/include/uapi/linux/limits.h
    // Unfortunetely, we can't just do the same with max path length itself.
    // https://eklitzke.org/path-max-is-tricky
    + 255;

    // inotify event max size formula:
    // sizeof(struct inotify_event) + NAME_MAX + 1
    // https://man7.org/linux/man-pages/man7/inotify.7.html
    std::mem::size_of::<inotify_sys::inotify_event>() + name_max + 1
}
```

epompeii:
This pull request increases the inotify event buffer size for the stream example from `32` to `1024` as discussed in Issue https://github.com/hannobraun/inotify-rs/issues/186.

Also included are two utility functions, `get_buffer_size()` and `get_absolute_path_buffer_size()`. The former is just a light, fallible wrapper around the latter. `get_buffer_size()` is meant to work with both relative and absolute paths, as it handles the canonicalization of the path. `get_absolute_path_buffer_size()` expects an absolute path to be passed in, and it is only meant to be used when a path is known to be absolute.

edit by @hannobraun:
Close #186 
(this should close the issue automatically when this is merged, so we don't forget)

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,6 +1,8 @@
 use std::{
+    io,
     mem,
-    os::unix::io::RawFd
+    os::unix::io::RawFd,
+    path::Path,
 };
 
 use inotify_sys as ffi;
@@ -9,6 +11,7 @@ use libc::{
     size_t,
 };
 
+const INOTIFY_EVENT_SIZE: usize = mem::size_of::<ffi::inotify_event>() + 257;
 
 pub fn read_into_buffer(fd: RawFd, buffer: &mut [u8]) -> isize {
     unsafe {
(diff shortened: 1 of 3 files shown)


src/util.rs
```diff
@@ -40,5 +43,53 @@ pub fn align_buffer_mut(buffer: &mut [u8]) -> &mut [u8] {
...
+/// See: [https://github.com/torvalds/linux/blob/master/include/uapi/linux/limits.h](https://github.com/torvalds/linux/blob/master/include/uapi/linux/limits.h)
+///
+/// Unfortunately, we can't just do the same with max path length itself.
+///
+/// See: [https://eklitzke.org/path-max-is-tricky](https://eklitzke.org/path-max-is-tricky)
+///
+/// This function is really just a fallible wrapper around `get_absolute_path_buffer_size()`.
+///
+/// path: A relative or absolute path for the inotify events.
+#[allow(dead_code)]
```
hannobraun, line 74: Why is `#[allow(dead_code)]` necessary? Since the function is public and re-exported from the crate, it shouldn't emit a warning.
```suggestion
```

src/util.rs
```diff
@@ -40,5 +43,53 @@ pub fn align_buffer_mut(buffer: &mut [u8]) -> &mut [u8] {
...
+    Ok(get_absolute_path_buffer_size(&path.canonicalize()?))
+}
+
+/// Get the inotify event buffer size for an absolute path
+///
+/// For relative paths, consider using `get_buffer_size()` which provides a fallible wrapper
+/// for this function.
+///
+/// path: An absolute path for the inotify events.
+#[allow(dead_code)]
```
hannobraun, line 85: Same as above.
```suggestion
```

hannobraun (approved):
Thank you, @epompeii!

I left two suggestions. Please let me know what you think.

src/util.rs
epompeii: Good catch! I think you are right.

src/util.rs
epompeii: Good catch! I think you are right.

epompeii:
Good catch! I think you are right.
I have made both changes.

hannobraun:
Thanks!

Issue #189, opened 2021-12-03
Using Inotify::read_events in async context


DBLouis:
Returning an empty iterator when `EGAIN/EWOULDBLOCK` is returned by `read` is not ideal in async context, for example in combination with tokio `AsyncFd`, because you need to detect "wouldblocks" but for that you have to consume the iterator. The iterator in neither `Clone` nor `ExactSizeIterator` which would help. I think that not handling this special case and letting the caller decide what to do with it is better.

hannobraun:
Hi @DBLouis, thank you for opening this issue!

So you're saying that [this code](https://github.com/hannobraun/inotify-rs/blob/e6d464de8b030ac3b8a53ffd912fdd09d14c9047/src/inotify.rs#L376-L381) should be removed, in favor of just returning the error? That sounds sensible to me, and I'd be happy to merge a pull request that does that.

DBLouis:
Thank you

hannobraun:
Addressed by #190.

PR #190, opened 2021-12-03
Return WouldBlock error as well in read_events

DBLouis:
#189


hannobraun:
Thank you for the pull request, @DBLouis!

Looks like we have some test failures. Looks like some tests relied on that behavior. Can you take a look?

hannobraun (approved):
Thank you, @DBLouis, looks great!

PR #196, opened 2022-08-11
Add Clippy lints to CI

### LINKED ISSUE #181: Consider using cargo clippy in test workflows
Clippy adds quite a few interesting lints that may be useful to have.

At the moment running clippy has 4 errors, mostly redundant field names in initialization. Clippy does also handle things such as warning you about unnecessary clones, redundant borrows on an already borrowed field, etc.

talklittle:
Fixes #181 

I also made an API deprecation, since Clippy was complaining about `Event.into_owned()` taking `&self` which is unexpected. `into_owned()` methods usually take `self` by value, not by reference. See https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention  
Resolution was to deprecate `into_owned()` and add the properly named method, `to_owned()`.  
This change may be relevant to #178

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,7 +12,7 @@ jobs:
     - name: Install rust
       uses: actions-rs/toolchain@v1
       with:
-        toolchain: 1.47.0
+        toolchain: 1.49.0
         override: true
     - name: Check rust and cargo version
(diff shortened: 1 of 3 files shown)


hannobraun (approved):
Thank you, @talklittle, looks great!

PR #198, opened 2022-08-15
Allow unaligned buffers

### LINKED ISSUE #174: Consider addressing buffer alignment in a more robust way
See #155 and and #171 (specifically https://github.com/hannobraun/inotify-rs/pull/171#pullrequestreview-655632788) for context.

talklittle:
The key is using [std::ptr::read_unaligned](https://doc.rust-lang.org/std/ptr/fn.read_unaligned.html) to read a struct from an unaligned pointer, instead of the `*` operator to dereference it. `read_unaligned()` permits unaligned pointers, whereas `*` leads to undefined behavior.

Adds a CI test to run [Miri](https://github.com/rust-lang/miri) on the affected code and detect alignment bugs. (Miri cannot be run on all tests, as it cannot make FFI calls for example.)

Fixes #174

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -29,3 +29,16 @@ jobs:
       run: cargo test --verbose --no-default-features
     - name: Run Clippy lints (no default features)
       run: cargo clippy --verbose --no-default-features -- -D warnings
+
+  miri:
+    name: "Miri"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Miri
+        run: |
+          rustup toolchain install nightly --component miri
+          rustup override set nightly
+          cargo miri setup
+      - name: Test with Miri tests with "from_buffer" in name
+        run: cargo miri test from_buffer
--- a/src/events.rs
+++ b/src/events.rs
@@ -12,7 +12,6 @@ use inotify_sys as ffi;
 
 use crate::fd_guard::FdGuard;
 use crate::watches::WatchDescriptor;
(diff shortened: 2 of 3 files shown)


.github/workflows/rust.yml
```diff
@@ -29,3 +29,16 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Miri
+        run: |
+          rustup toolchain install nightly --component miri
+          rustup override set nightly
+          cargo miri setup
+      - name: Test with Miri tests with "from_buffer" in name
+        run: cargo miri test from_buffer
```
hannobraun, line 44: Might be worth running Miri for all tests, just in case. Or is there a specific reason not to?

hannobraun (approved):
Thank you, @talklittle, looks awesome!

I left one comment with a question/suggestion. This can be addressed later though (or not at all), so I'm merging this pull request right now.

.github/workflows/rust.yml
talklittle: Yes, unfortunately Miri errors out on many of the tests and doctests. For example `TestDir::new()` depends on `libc::clock_gettime()` which gives the error "`clock_gettime` not available when isolation is enabled". Miri cannot do FFI calls either, so it errors on anything calling `ffi::inotify_init()`: "can't call foreign function: inotify_init".

hannobraun: Ah, got it. Too bad, what can we do!

PR #199, opened 2022-08-22
Replace event_stream() with into_event_stream()

### LINKED ISSUE #176: `Inotify::event_stream` is error-prone
`Inotify::event_stream` enabled concurrent use of one `Inotify` instance and an arbitrary number of `EventStream` instances, which is error-prone. It should be replaced with an `into_event_stream` method that consumes the `Inotify`. Ideally, there should also be a matching method on `EventStream`, to convert back to `Inotify`.

See https://github.com/hannobraun/inotify-rs/issues/112#issuecomment-850352465 and https://github.com/hannobraun/inotify-rs/issues/112#issuecomment-850418244.

talklittle:
Also adds EventStream.into_inotify() to convert back
to an Inotify.

event_stream() was problematic because it could allow
a caller to inadvertently create multiple streams
reading and contending over the same source, resulting
in unpredictable event emission.

Fixes #176

### DIFF
--- a/examples/stream.rs
+++ b/examples/stream.rs
@@ -14,7 +14,7 @@ use tempfile::TempDir;
 
 #[tokio::main]
 async fn main() -> Result<(), io::Error> {
-    let mut inotify = Inotify::init()
+    let inotify = Inotify::init()
         .expect("Failed to initialize inotify");
 
     let dir = TempDir::new()?;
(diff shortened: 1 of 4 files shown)


hannobraun (approved):
Thank you, @talklittle. Looks great!

PR #202, opened 2023-06-06
Implement `AsFd` and bidirectional conversion to/from `OwnedFd`

joshtriplett:
In particular, this makes it possible to get a `BorrowedFd` from an
`Inotify`, which allows integration with polling mechanisms that don't
use `RawFd`. This will be needed to work with the next version of
`async-io`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,7 +12,7 @@ jobs:
     - name: Install rust
       uses: actions-rs/toolchain@v1
       with:
-        toolchain: 1.49.0
+        toolchain: 1.63.0
         override: true
         components: clippy
     - name: Check rust and cargo version
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ authors = [
   "Frank Denis <github@pureftpd.org>"
 ]
 edition = "2018"
+rust-version = "1.63"
 
 description   = "Idiomatic wrapper for inotify"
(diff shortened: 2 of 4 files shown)


hannobraun:
Thank you for the pull request, @joshtriplett! This looks good, but CI fails because our MSRV is too low. I've opened #203.

I'll probably find some time to address this soon, but if anyone's reading this while I still haven't, feel free to jump in with a pull request! This PR is blocked until then.


> This will be needed to work with the next version of `async-io`.

I assume this means you need me to publish a new release, once this is merged?

joshtriplett:
@hannobraun Right, this PR itself will raise the MSRV as well, for `AsFd`/`BorrowedFd`/`OwnedFd`.

joshtriplett:
@hannobraun I've updated this to also fix the MSRV and CI. 1.63, which has the I/O Safety types stabilized, will ship in the Debian stable release shipping in a few days.

hannobraun:
Thank you, @joshtriplett! It would have been a bit nicer go have the MSRV bump in a separate pull request, so I don't get confused later when writing the changelog. But this works too!

> And yes, I'd really appreciate a release with this change, once merged.

Understood. I'll try to get it out today.

hannobraun:
New version is out: https://github.com/hannobraun/inotify-rs/pull/204, https://crates.io/crates/inotify/0.10.1

joshtriplett:
@hannobraun Thank you!

PR #213, opened 2024-02-09
docs: link to rust items by name

### LINKED ISSUE #212: docs.rs links from inotify to inotify_sys are broken
The docs for each of the [`WatchMask`](https://docs.rs/inotify/0.10.2/inotify/struct.WatchMask.html) and [`EventMask`](https://docs.rs/inotify/0.10.2/inotify/struct.EventMask.html) variants try to link to the docs.rs page for `inotify_sys`, but the links are all broken.

For example:
https://docs.rs/inotify/latest/inotify_sys/constant.IN_ACCESS.html is 404.

It should be:
https://docs.rs/inotify-sys/latest/inotify_sys/constant.IN_ACCESS.html

eric-seppanen:
- For items that are in scope, remove the explicit html path.
- For items not in scope, link to them by a rust path that does work (e.g. `crate::Watches` or `libc::read`).
- Fix incorrect reference `WatchMask::IGNORED` -> `EventMask::IGNORED`.
- Replace verbose `[https://some-url/](https://some-url/)` with the equivalent `<https://some-url/>`.

This fixes many broken links (some always broken, some only broken on docs.rs), and makes them more robust in the future (rustdoc emits warnings most items that don't resolve.)

Fixes #212.

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -19,8 +19,8 @@ use crate::watches::WatchDescriptor;
 /// Allows for iteration over the events returned by
 /// [`Inotify::read_events_blocking`] or [`Inotify::read_events`].
 ///
-/// [`Inotify::read_events_blocking`]: struct.Inotify.html#method.read_events_blocking
-/// [`Inotify::read_events`]: struct.Inotify.html#method.read_events
+/// [`Inotify::read_events_blocking`]: crate::Inotify::read_events_blocking
+/// [`Inotify::read_events`]: crate::Inotify::read_events
 #[derive(Debug)]
 pub struct Events<'a> {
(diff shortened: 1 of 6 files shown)


hannobraun (approved):
Thank you, @eric-seppanen, looking good!

I only took a quick look at at the diff and didn't actually test any links. It's getting late here, and if I don't review this now, I'll probably not get to it until some time next week. But whether everything works perfectly or not, this is definitely taking things in the right direction, so I'm happy to merge!

hannobraun:
It would be good to prevent this from happening again in the future. I've opened https://github.com/hannobraun/inotify-rs/issues/214.

PR #224, opened 2024-09-17
Format README example

allan2:
This runs rustfmt on the README example.
I also removed `extern` as it is rarely used since Rust 2018.


hannobraun (approved):
Thank you, @allan2!

PR #228, opened 2025-01-06
Add async doc in top doc

jokemanfire:
add doc introduction for async stream.


hannobraun (approved):
This is great, @jokemanfire! Thank you!

PR #230, opened 2025-01-14
Drop ArcFdGuard newtype

WhyNotHugo:
This type existed because AsRawFd was not implemented for `Arc<T> where T: AsRawFd`.

As for 1.63.0, the following implementation is provided:

    impl<T: AsRawFd> AsRawFd for Arc<T>

The current MSRV is 1.70, so we can rely on this new implementation.

Drop the newtype wrapper, and rely on the new stdlib implementation instead.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -1,6 +1,6 @@
 use std::{
     io,
-    os::unix::io::{AsRawFd, RawFd},
+    os::unix::io::AsRawFd,
     pin::Pin,
     sync::Arc,
     task::{Context, Poll},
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
I have a follow-up for this which replaces FdGuard with OwnedFd: https://github.com/hannobraun/inotify-rs/commit/ad90da8bc9009a42192ee746a4665da532d4538a

hannobraun (approved):
Thank you, @WhyNotHugo! Looks good!

PR #231, opened 2025-01-14
Skip testing code which doesn't compile

WhyNotHugo:
This sample code isn't standalone, and relies on undeclared variables. Skip it, so that `cargo test` doesn't fail.


hannobraun (approved):
Thank you, @WhyNotHugo!

It's weird, that the CI build doesn't catch this. I've opened https://github.com/hannobraun/inotify-rs/issues/232.

PR #235, opened 2025-04-20
refactor buffer size calculation

JackDyre:
Improve the readability of the buffer size calculation.


hannobraun (approved):
Thank you, @JackDyre, looks great!

PR #236, opened 2025-04-22
Parsed Repesentation of Event Masks

### LINKED ISSUE #76: Review representation of `mask` in `Event`
From reading the man pages, I get the impression that distinct event objects are generated for each type of event, and that therefore most bits in the event mask can't coexist. If that is the case, maybe it makes sense to create a new enum, `EventKind`, to represent the type of event.

One exception to this case is `IN_ISDIR`, which can be set in addition to other bits that identify the type of event. A new `bool` field could be created specifically for this flag.

In any case, before anything like this can be merged, more careful reading of the man pages and probably some experimentation is required.

JackDyre:
resolves #76 

Adds types to represent event masks in a structured way. To preserve backward compatibility, it is "opt-in", meaning users have to manually call the `EventMask::parse()` method to get this structured representation of the event masks, meaning all existing code that depends on this library should be unaffected.

I opted to represent the auxiliary flags as a struct of `bool`s rather than a `bitflags!` instance. I likely should add some helper getter methods on the `ParsedEventMask` type to make it easier to access the inner auxiliary fields.

Addresses https://github.com/hannobraun/inotify-rs/issues/76#issuecomment-335412259
Currently does not address https://github.com/hannobraun/inotify-rs/issues/76#issuecomment-335754902

`ParsedEventMask` is the same size as `EventMask` (if Rust allowed <1byte types it could all fit in a u8 but oh well)

#### This PR still needs to address:
- [x] Tests
- [x] Trait impls (eq, etc...)
- [x] Possibly store context in the Error type

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -348,7 +348,7 @@ impl EventMask {
 
 /// A struct that provides structured access to event masks
 /// returned from reading an event from an inotify fd
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct ParsedEventMask {
     /// The kind of event that occurred
     pub kind: Option<EventKind>,
@@ -390,7 +390,7 @@ impl TryFrom<EventMask> for ParsedEventMask {
 ///
 /// Exactly 0 or 1 of these bitflags will be set in an event mask
 /// returned from reading an inotify fd
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum EventKind {
     /// File was accessed (e.g., [`read(2)`], [`execve(2)`])
     ///
@@ -513,7 +513,7 @@ impl TryFrom<EventMask> for Option<EventKind> {
 /// The non-mutually-exclusive bitflags that may be set
 /// in an event read from an inotify fd. 0 or more of these
 /// bitflags may be set.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
 pub struct EventAuxiliaryFlags {
     /// Watch was removed when explicitly removed via [`inotify_rm_watch(2)`]
     /// or automatically (because the file was deleted or the filesystem was unmounted)
@@ -548,7 +548,7 @@ impl From<EventMask> for EventAuxiliaryFlags {
 }
 
 /// An error that occured from parsing an raw event mask
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum EventMaskParseError {
     /// More than one bit repesenting the event type was set
     TooManyBitsSet,
--- a/src/events.rs
+++ b/src/events.rs
@@ -330,7 +330,7 @@ bitflags! {
 
 impl EventMask {
     /// Parse this event mask into a ParsedEventMask
-    pub fn parse(self: Self) -> Result<ParsedEventMask, EventMaskParseError> {
+    pub fn parse(self) -> Result<ParsedEventMask, EventMaskParseError> {
         self.try_into()
     }
 
--- a/src/events.rs
+++ b/src/events.rs
@@ -352,16 +352,16 @@ impl EventMask {
 pub struct ParsedEventMask {
     /// The kind of event that occurred
     pub kind: Option<EventKind>,
-    /// The auxillary flags about the event
-    pub auxillary_flags: EventAuxillaryFlags,
+    /// The auxiliary flags about the event
+    pub auxiliary_flags: EventAuxiliaryFlags,
 }
 
 impl ParsedEventMask {
     /// Construct a `ParsedEventMask` from its component parts
-    pub fn from_parts(kind: Option<EventKind>, auxillary_flags: EventAuxillaryFlags) -> Self {
+    pub fn from_parts(kind: Option<EventKind>, auxiliary_flags: EventAuxiliaryFlags) -> Self {
         ParsedEventMask {
             kind,
-            auxillary_flags,
+            auxiliary_flags,
         }
     }
 
@@ -372,9 +372,9 @@ impl ParsedEventMask {
         }
 
         let kind = mask.try_into()?;
-        let auxillary_flags = mask.into();
+        let auxiliary_flags = mask.into();
 
-        Ok(ParsedEventMask::from_parts(kind, auxillary_flags))
+        Ok(ParsedEventMask::from_parts(kind, auxiliary_flags))
     }
 }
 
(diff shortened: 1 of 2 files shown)


JackDyre:
Oops I just realized I misspelled auxiliary in all of my code. I will definitely fix that

src/events.rs
```diff
@@ -338,6 +346,229 @@ impl EventMask {
...
+        });
+
+        // Optionally take the first matching bitflag
+        let kind = kinds.next();
+
+        if kinds.next().is_some() {
+            // The mask is invalid.
+            //
+            // More than one of the bitflags are set
+            return Err(EventMaskParseError::TooManyBitsSet);
```
hannobraun, line 496: I wonder if this should be an `unreachable!` panic instead of an error. But it's probably better to not make too many assumptions about the Linux kernel, and leave this as-is.

hannobraun (commented):
Thank you for opening this pull request, @JackDyre! I did a quick review, and this looks great!

I'd happy to merge this, pending a closer review (which I didn't do yet, as you marked this as draft). ~~What would you like to change, before you think this is ready?~~

hannobraun:
Ah, sorry for the stupid question in my review message. I read your checklist of open issues initially, then promptly forgot about it again :joy:

src/events.rs
JackDyre: I'm not sure if a panic is desirable. There is no guarantee that the `EventMask` being parsed came directly from the kernel. The user could mutate the mask before parsing, or since it is a public type, the user could construct their own for some reason.

I'm not sure if an error is the correct choice though, it's worth considering just taking the first matching bitflag and moving on (and then documenting that if for some reason multiple bits are set, there is no guarantee across versions which one gets chosen).

hannobraun: > I'm not sure if a panic is desirable. There is no guarantee that the EventMask being parsed came directly from the kernel. The user could mutate the mask before parsing, or since it is a public type, the user could construct their own for some reason.

Good points!

> I'm not sure if an error is the correct choice though, it's worth considering just taking the first matching bitflag and moving on (and then documenting that if for some reason multiple bits are set, there is no guarantee across versions which one gets chosen).

No, let's keep it as an error. Otherwise we'd be hiding bugs or wrong assumptions, and that only makes them more painful to deal with.

src/events.rs
```diff
@@ -338,6 +346,229 @@ impl EventMask {
...
+}
+
+impl ParsedEventMask {
+    /// Construct a `ParsedEventMask` from its component parts
+    pub fn from_parts(kind: Option<EventKind>, auxillary_flags: EventAuxillaryFlags) -> Self {
+        ParsedEventMask {
+            kind,
+            auxillary_flags,
+        }
+    }
```
hannobraun: This constructor seems redundant, with all fields being public anyway. Not really a problem, but I don't think it's necessary either.

src/events.rs
```diff
@@ -338,6 +346,229 @@ impl EventMask {
...
+
+    /// Parse a raw event mask
+    pub fn from_raw_event_mask(mask: EventMask) -> Result<Self, EventMaskParseError> {
+        if mask.contains(EventMask::Q_OVERFLOW) {
+            return Err(EventMaskParseError::QueueOverflow);
+        }
+
+        let kind = mask.try_into()?;
+        let auxillary_flags = mask.into();
```
hannobraun: _Very_ minor nitpick: These `into` and `try_into` calls can sometimes make code a bit hard to follow, since it's not always immediately obvious what type the value is being converted into. In general, unless when using an `impl Into<...>` parameter, I prefer to make it explicit using something like `EventAuxiliaryFlag::from(mask)`.

src/events.rs
```diff
@@ -338,6 +346,229 @@ impl EventMask {
...
+
+    fn try_from(value: EventMask) -> Result<Self, Self::Error> {
+        Self::from_raw_event_mask(value)
+    }
+}
+
+/// Represents the type of inotify event
+///
+/// Exactly 0 or 1 of these bitflags will be set in an event mask
+/// returned from reading an inotify fd
```
hannobraun, line 392: I think this is slightly confusing for someone not knowing the full context, as it refers to bitflags, when this is an enum. Might be better to remove it here and instead add this information to the `kind: Option<EventKind>` field instead. Specifically the `0 or 1` part makes more sense there, with the `Option` directly beside it.

src/events.rs
```diff
@@ -338,6 +346,229 @@ impl EventMask {
...
+    /// attributes ([`setxattr(2)`]), link count (since Linux
+    /// 2.6.25; e.g., for the target of [`link(2)`] and for
+    /// [`unlink(2)`]), and user/group ID (e.g., [`chown(2)`])
+    ///
+    /// [`chmod(2)`]: https://man7.org/linux/man-pages/man2/chmod.2.html
+    /// [`utimensat(2)`]: https://man7.org/linux/man-pages/man2/utimensat.2.html
+    /// [`setxattr(2)`]: https://man7.org/linux/man-pages/man2/setxattr.2.html
+    /// [`link(2)`]: https://man7.org/linux/man-pages/man2/link.2.html
+    /// [`unlink(2)`]: https://man7.org/linux/man-pages/man2/unlink.2.html
+    /// [`chown(2)`]: https://man7.org/linux/man-pages/man2/chown.2.html
```
hannobraun, line 412: Appreciate the references here and in the other doc comments. Awesome! :+1:

src/events.rs
```diff
@@ -338,6 +346,229 @@ impl EventMask {
...
+        match self {
+            Self::TooManyBitsSet => {
+                writeln!(f, "Error parsing event mask: too many event type bits set")
+            }
+            Self::QueueOverflow => writeln!(f, "Error: the kernel's event queue overflowed"),
+        }
+    }
+}
+
+impl Error for EventMaskParseError {}
```
hannobraun: Should we pull [`thiserror`](https://crates.io/crates/thiserror) in, to take care of this kind of stuff a bit more conveniently? Might not be worth it, since with such a low-level library, fewer dependencies are always better.

hannobraun (approved):
Thank you again, @JackDyre, awesome work!

I left a few comments; most very nitpicky, none of them important. Feel free to follow up on them or not, as you think is best. Either way, I'm going to merge this now, since it's already great!

src/events.rs
JackDyre: It's up to you.

I feel like `thiserror` is so common, it likely wouldn't be a new dependency for most of `inotify-rs`'s dependents. (I wonder if there is a tool out there that can analyze dependents on crates.io for this purpose).

I'm willing to open a PR to add it if you'd like.

src/events.rs
JackDyre: I can remove it if you'd like. It's just a pattern I use because I personally prefer the nvim code completion snippet for method calls over the one for struct literals

src/events.rs
JackDyre: I would say pulling `thiserror` as a dep would definitely be worth it if you intend on developing a general `InotifyError` that handles a more error cases throughout the library.

hannobraun: I don't feel strongly about it, one way or the other. Maybe let's leave it out for now (this code is already written, after all), but keep it in mind for future additions.

src/events.rs
hannobraun: Let's leave it in then. It's only a tiny burden, if at all. If it's even a tiny bit helpful, that's probably worth it.

PR #238, opened 2025-04-25
Adjustments to parsed event mask

JackDyre:
Adjustments to #236

See https://github.com/hannobraun/inotify-rs/pull/236#discussion_r2056667361 and https://github.com/hannobraun/inotify-rs/pull/236#discussion_r2056671559


hannobraun (approved):
Thank you for the follow-up, @JackDyre. Looks great!

PR #239, opened 2025-04-27
fix doc code compilation

### LINKED ISSUE #232: CI build doesn't compile examples in documentation
One example is currently broken on `main`. If I run `cargo test` locally, I get this:

```
test src/lib.rs - (line 54) ... FAILED
```

(See https://github.com/hannobraun/inotify-rs/pull/231, which fixes this.)

And yet, the CI build is green right now: https://github.com/hannobraun/inotify-rs/actions/runs/12631159244/job/35192359287

Inspecting the output, it seems that doc tests aren't run. And indeed, if I run the same command locally (`cargo test --all-features --all-targets`), I'm getting the same result: all green.

I don't know why that is. It would be great, if somebody could look into this, and figure out a better command for the CI build to run.

JackDyre:
resolves #232

- removed the `skip` marker
- added import for the `.next()` method
- wrapped it in a hidden async fn for `.await`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -51,12 +51,22 @@
 //! ```
 //! When you want to read events asynchronously, you need to convert it to [`EventStream`].
 //! The transform function is [`Inotify::into_event_stream`]
-//! ```skip
-//!  let mut stream = inotify.into_event_stream(&mut buffer)?;
-//!  // Read events from async stream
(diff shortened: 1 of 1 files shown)


hannobraun:
I've edited the PR description to prevent it from closing the issue. Merging now. Thanks again!

PR #246, opened 2025-10-16
Implement EventStream::read_events

WhyNotHugo:
`EventStream::read_events` allows reading events from an `EventStream` without blocking (or awaiting).

This is useful to drain the queue of all immediately pending events.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -46,6 +46,72 @@ where
         Watches::new(self.fd.get_ref().clone())
     }
 
+    /// Reads and returns a single available event
+    ///
+    /// Returns the next available event from the internal buffer. If the buffer
+    /// is empty, it will attempt to read more events from the file descriptor.
+    /// Returns `Ok(None)` if no events are available without blocking.
+    ///
+    /// Please note that inotify will merge identical successive unread events
+    /// into a single event. This means this method can not be used to count the
+    /// number of file system events.
+    ///
+    /// # Errors
+    ///
+    /// This function directly returns all errors from the call to [`read`],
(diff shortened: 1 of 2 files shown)


hannobraun (commented):
Thank you for this pull request, but I'm not convinced this belongs here. `EventStream` already implements `Stream`, and this seems like something that should be possible based on that generic primitive.

Maybe something like this (untested):

```rust
let stream = event_stream.peekable();

while let Some(peek) = stream.peek().now_or_never() {
    let _ = stream.next();
}
```

That would be using [`StreamExt::peekable`](https://docs.rs/futures/latest/futures/prelude/stream/trait.StreamExt.html#method.peekable), [`Peek::peek`](https://docs.rs/futures/latest/futures/stream/struct.Peekable.html#method.peek), [`FutureExt::now_or_never`](https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.now_or_never), and [`StreamExt::next`](https://docs.rs/futures/latest/futures/prelude/stream/trait.StreamExt.html#method.next).

Unless you can come up with a convincing argument for why this functionality needs to be here specifically, I'm inclined to close this pull request.

WhyNotHugo:
[FutureExt::now_or_never](https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.now_or_never) only returns if there are events in the buffer, but not if events are ready to be read immediately without blocking.

WhyNotHugo:
I wrote a standalone test in https://github.com/hannobraun/inotify-rs/pull/248 to property assert his behaviour, and `now_or_never` works fine :)

PR #248, opened 2025-11-30
Test draining without blocking

WhyNotHugo:
Ensure that the queue can be drained by reading all pending events without ever blocking (and performing only non-blocking reads).

Closes: https://github.com/hannobraun/inotify-rs/pull/246


WhyNotHugo:
v2: remove unused import when feature is disabled

PR #257, opened 2026-03-17
Replace futures-core with futures-util to enable StreamExt re-exports

### LINKED ISSUE #253: Consider re-exporting `futures_util::StreamExt`
When using this crate's async API, the `Inotify` instance needs to be converted into an `EventStream`.
Accessing that stream requires the `next()` method from `futures_util::StreamExt` according to the example.
Because `StreamExt` is not re‑exported, consumers of this crate must add `futures_util`,
with a matching version, to their own Cargo.toml to be able to use it.

Please consider re‑exporting the `StreamExt` trait so that users can access the async API without adding extra dependencies.
As an alternative, please provide an example how to use it without additional application-side dependencies.
This avoids adding `futures_util` as a dependency to this crate.

maximilian-maisel-bl:
Switching to futures-util allows re‑exporting the StreamExt trait and using the crates async APIs without requiring an additional downstream dependency.

Resolves #253

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,12 +23,12 @@ exclude    = ["/inotify-rs.sublime-project"]
 
 [features]
 default = ["stream"]
-stream = ["futures-core", "tokio"]
+stream = ["futures-util", "tokio"]
 
 
 [dependencies]
 bitflags     = "2"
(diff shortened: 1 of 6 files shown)


hannobraun (approved):
Thank you, @maximilian-maisel-bl, looks great!

PR #258, opened 2026-05-08
 Support FreeBSD kcmp in fd guard equality 

kev009:
Forwards to `inotify-sys/freebsd-native` so consumers can build the inotify crate against FreeBSD 14.4+'s in-kernel inotify(2) (libc wrappers) instead of the userspace libinotify shim.

The high-level crate already uses only portable std::os::unix and libc { fcntl, F_GETFL, F_SETFL, O_NONBLOCK }, so no source changes are required to compile on FreeBSD.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -47,6 +47,23 @@ jobs:
       - name: Run tests (no default features)
         run: cargo test --no-default-features --all-targets --verbose
 
+  test-freebsd:
+    name: Test (FreeBSD)
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+          shell: cpa.sh {0}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Test on FreeBSD
+        uses: cross-platform-actions/action@v1.3.0
+        with:
+          operating_system: freebsd
+          version: "15.1"
+      - run: |
+          sudo pkg install -y rust
+          cargo test
+
   miri:
     name: "Miri"
     runs-on: ubuntu-latest
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,11 +19,10 @@ exclude    = ["/inotify-rs.sublime-project"]
 default = ["stream"]
 stream = ["futures-util", "tokio"]
 
-
 [dependencies]
 bitflags     = "2"
 futures-util = { version = "0.3.30", optional = true }
-inotify-sys  = "0.1.7"
+inotify-sys  = "0.1.8"
 libc         = "0.2"
 tokio        = { version = "1.40.0", optional = true, features = ["net"] }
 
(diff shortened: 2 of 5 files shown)


hannobraun (changes requested):
Thank you, @kev009! I'm going to mark this as a draft, as this can't really work until https://github.com/hannobraun/inotify-sys/pull/26 made its way into a release. When that has happened, we should also update the dependency here to the new version, as I don't think it would work, if it resolved to an older one.

hannobraun:
The upstream change is now released: https://crates.io/crates/inotify-sys/0.1.7

hannobraun (changes requested):
Please update the dependency on `inotify-sys` to `0.1.7`. As it is, Cargo might resolve to an older version that doesn't have this feature.

I'm also wondering, if we need CI coverage for this. Maybe not, since this just passes through a feature to the `inotify-sys` crate. Though, as I've realized now, the CI build for `inotify-sys` doesn't cover this feature either.

I don't know. I guess I won't insist, since the change over here is minimal. But I'll probably be more strict about the CI coverage of subsequent FreeBSD-related changes, to limit the potential damage.

kev009:
Pushed, there seemed to be an unrelated build break I propose a fix in the second commit.

Re testing.. when I opened this we couldn't assume that inotify was available.  But now all supported versions of FreeBSD ship inotify, so should we consider dropping the feature guard, so the CI always uses it?

hannobraun (changes requested):
Thank you, @kev009!

Your new commit looks good, but crosses the line of what I'm willing to accept without CI coverage. To get this merged, I'll need the CI build (including the test suite) running on FreeBSD.

Since there's already [prior art in `inotify-sys`](https://github.com/hannobraun/inotify-sys/blob/ba9ab1fc6473d5de0b7d903a757924a8d84bf5d4/.github/workflows/ci.yml) this shouldn't be too hard to do, I hope.

> Re testing.. when I opened this we couldn't assume that inotify was available. But now all supported versions of FreeBSD ship inotify, so should we consider dropping the feature guard, so the CI always uses it?

Sounds reasonable, and fewer moving parts are definitely preferable. But I think this shouldn't be my decision. If you, the top FreeBSD expert in this pull request, tell me that this is the right call, then that's what I'll merge.

If you think that this may be controversial, you could try to build consensus with [other interested parties](https://github.com/hannobraun/inotify-sys/issues/24). But either way, I don't know much about FreeBSD, so I'm happy to merge whatever looks reasonable and passes the test suite.

kev009:
Ok, sent updates.  Requires, presumably, inotify-sys 0.1.8.

hannobraun:
Thank you, @kev009. This looks good at a glance. Let's continue here, once https://github.com/hannobraun/inotify-sys/pull/32 is merged and inotify-sys 0.1.8 is out.

hannobraun:
inotify-sys v0.1.8 is out now: https://github.com/hannobraun/inotify-sys/pull/33

kev009:
@hannobraun rebased

hannobraun (changes requested):
Thank  you, @kev009! Lots of test failures though.

kev009:
@hannobraun resolved in a way that is hopefully more correct.

hannobraun (approved):
Awesome, that worked. Thank you, @kev009!

PR #260, opened 2026-05-29
Reopen PR (use kcmp to compare watch descriptors from duplicated fds) #251

khanhtranngoccva:


hannobraun (approved):
Thank you, @khanhtranngoccva! Looks good!

PR #263, opened 2026-07-04
Bump inotify-sys to 0.1.7, add `WatchMask::MASK_CREATE` option.

CircuitCoder:
This PR bumps the version of `inotify-sys` to 0.1.7, which contains the `IN_MASK_CREATE` constant.

It also added a bitflag `MASK_CREATE` to WatchMask, mirroring that constant.

`inotify-sys` 0.1.7 also contain the `freebsd-native`. There seems to be ongoing work with FreeBSD builds in #258, which also bumps `inotify-sys`. Maybe I should wait after that's merged, and then rebase onto it?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ stream = ["futures-util", "tokio"]
 [dependencies]
 bitflags     = "2"
 futures-util = { version = "0.3.30", optional = true }
-inotify-sys  = "0.1.5"
+inotify-sys  = "0.1.7"
 libc         = "0.2"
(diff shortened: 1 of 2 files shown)


hannobraun (approved):
Thank you, @CircuitCoder, looks good!

> `inotify-sys` 0.1.7 also contain the `freebsd-native`. There seems to be ongoing work with FreeBSD builds in #258, which also bumps `inotify-sys`. Maybe I should wait after that's merged, and then rebase onto it?

That pull request is still in progress and is going to require another release of `inotify-sys`. Since this one is ready now, I'll merge it first.

